PR #1, opened 2020-06-15
Upgrade kernel from 4.14 to 5.6

bstreiff:
This is a rebase of our kernel commits on top of [linux-rt-devel/v5.6.17-rt9](https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-rt-devel.git/tag/?h=v5.6.17-rt9). This review should include most NI out-of-tree changes relevant to x64 platforms (excepting those already noted as requiring significant rework).

- [x] Boot tested on cRIO-9030
- [x] Verified that major peripherals like ethernet and serial work.
- [x] Run cyclictest + hackbench to verify that there are no major performance regressions.

[procedural note: For convenience (since v5.6.17-rt9 is the base for this review) [here's the diff of arch/x86/configs/nati_x86_64_defconfig between our 4.14 and this branch](https://gist.github.com/bstreiff/b11195498c1902949a864443f9ffa1e2) generated with `git diff nilrt/master/4.14 dev/rebase/nilrt/master/5.6 -- arch/x86/configs/nati_x86_64_defconfig`. We should keep comments on the defconfig changes in this PR though.]
[also procedural note: I inadvertently called the base branch 'linux-stable-rt' when it actually came out of 'linux-rt-devel']


bstreiff:
courtesy ping for other people still at NI with GitHub accounts that have commits in here: @rtollert @jeminor @xhuff @kroeschl @harisokanovic @chenyeechew @billpittman

gratian (commented):
Commit 99820a77cff0 (" kernel: Providing API to allow userland programs to request a cold or warm reboot") suggests the current implementation could be replaced with an EFI reboot which would eliminate the need for an out of tree patch. We should consider tracking that as a separate work item.

gratian:
Consider dropping both these commits:
edbf774948cd ("Revert "serial/8250: Add support for NI-Serial PXI/PXIe+485 devices")
440a996b0aff ("serial/8250: Add support for NI-Serial PXI/PXIe+485 devices")

Since both the commit and revert are part of the same rebase there's no reason to pollute the history here.

gratian:
If I recall correctly commit e22a5387f311 ("nati_x86_64_defconfig: Use the "guess" stack unwinder by default") was needed due to an older OpenEmbedded toolchain not supporting the better/newer ORC stack unwinder. Consider creating a work item to track switching the stack unwider after the OS/toolchain upgrade is complete.

gratian:
After commit f0ec643636a4 ("Revert "time: Always make sure wall_to_monotonic isn't positive"") the 'ret' local variable in do_settimeofday64() is not assigned to after initialization and is also not used as the return variable anymore. However is still used in an 'if' condition:
```
        if (!ret)
                audit_tk_injoffset(ts_delta);
 
-       return ret;
+       return 0;
```
Consider removing it altogether as part of the revert.

gratian:
@jeminor Can you comment if the WEXT interface is still used with the wl12xx driver? I.e. is commit 17c9850f8056 ("cfg80211: wext: Force scans to occur in AP mode") still required?

gratian:
Consider cleaning up the 'rar_count' variable added by commit a1948ae015b5 ("e1000e: Add delays after writing to registers"). It doesn't appear to be in use anymore.

gratian:
I believe the following commits:

f28ef09d9e17d5f39f2ee4bcebae516672881515 driver core: NI backwards compatibility for device_poll interface
bfddf501aa6eeadd84a08aefef0aaa685329dc80 net: add device_poll functionality common to all net devices
a23e50e51fcd9fe1ce0ff2aea38d918dd9c37129 driver core: add device_poll interface

where only ever used on ARM/Zynq targets and they are not needed for x86_64. Turns out we did not get significant RT performance improvements from implementing polling vs. interrupt based networking on x86_64 so it was not implemented there.

This is further confirmed by the fact that the config options added by these commits:

- CONFIG_DEVICE_POLL
- CONFIG_DEVICE_POLL_NI_COMPAT
- CONFIG_NETDEV_POLL

are only present in the ARM/Zynq config and are not enabled in the x86_64 case.

Consider dropping these commits from the x86_64 rebase.

bstreiff:
IIRC the wireless configuration pieces that roll up into `nitargetcfg` still use the wext ioctls in some places and hadn't been updated to use nl80211, so I don't think we can drop that one.

bstreiff:
Dropped:
- edbf774 ("Revert "serial/8250: Add support for NI-Serial PXI/PXIe+485 devices")
- 440a996 ("serial/8250: Add support for NI-Serial PXI/PXIe+485 devices")
- f28ef09 driver core: NI backwards compatibility for device_poll interface
- bfddf50 net: add device_poll functionality common to all net devices
- a23e50e driver core: add device_poll interface

Cleaned up codepaths in f0ec643 ("Revert "time: Always make sure wall_to_monotonic isn't positive"") and a1948ae ("e1000e: Add delays after writing to registers") as recommended.

gratian:
LGTM. Ship it!

Acked-by: Gratian Crisan \<gratian.crisan@ni.com\>

bstreiff:
Pushed as [nilrt/master/5.6](https://github.com/ni/linux/tree/nilrt/master/5.6).

(Procedural note: I'm not completing it as a PR in the github workflow because the PR is against a synthetic v5.6.17-rt9 branch.)

PR #3, opened 2020-08-05
futex: Add work-around for inconsistent pi_state

gratian:
Commit 73d786bd043e ("futex: Rework inconsistent rt_mutex/futex_q
state") removed a work-around in fixup_owner() that fixed an
inconsistent pi_state.

Part of the work-around was added back by commit 1352130fe6aa ("futex:
Avoid violating the 10th rule of futex"). However it left out, for
good reason (it might hide a race), the fix-up for the case where the
rt_mutex ends up w/o an owner.

This commit adds back the fix-up for the pi_state pi_mutex owner to
work around the race that is still present in the kernel. This allows
us to recover and continue if we find an rt_mutex w/o an owner. We
also throw a warning so we can continue the search for the race.

### DIFF
--- a/kernel/futex.c
+++ b/kernel/futex.c
@@ -2448,6 +2448,16 @@ retry:
 		 * Since we just failed the trylock; there must be an owner.
 		 */
 		newowner = rt_mutex_owner(&pi_state->pi_mutex);
+		WARN_ON(!newowner);
+
+		/*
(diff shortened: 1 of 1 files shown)


bstreiff (approved):

gratian:
Merged as commit 4fe2aca61b3001a9ef3adf1b6179de239ba16124 "futex: Add work-around for inconsistent pi_state"

PR #4, opened 2020-08-19
Fix build of 4.14 kernel in dunfell OE

chaitu236:
dunfell OE sets a gcc parameter for CFLAGS which is inherited by this Makefile. If host's gcc is old and doesn't support the parameter, build fails.

Cherry-picked this commit from upstream 4.14 linux-stable-rt that fixes it.


chaitu236:
@ni/rtos

gratian (approved):
Ship it!

PR #5, opened 2020-08-28
nati_x86_64_defconfig: Increase NR_CPUS to 64

gratian:
Increase the maximum number of CPUs supported by the kernel to 64.
This allows us to fully support new hardware at the cost of a slight
increase in memory usage: ~8KB/CPU according to the kernel help
documentation.


gratian:
Note for reviewers: the diff looks wonky because the x86_64 default is 64 and I just bumped it to that so it makes it look like I've removed that line altogether.

bstreiff:
If this makes for a wonky diff, maybe the tone of the change description should be "go back to default NR_CPUS instead of limiting it to 16"?

gratian:
Good idea on the description change. I will do that before pushing (I don't think I can do that as a fixup commit).

bstreiff:
ACK on the change itself, just needs a better description.

bstreiff (approved):

gratian:
Pushed as commit 997fc3df5a3e740f585e6298cae9bd9d63832f7a ("nati_x86_64_defconfig: Revert NR_CPUS to the default 64")

PR #11, opened 2020-12-18
nati_zynq_defconfig: Fix typo error in commit 9e8c398

NIwkoe:
The previous commit was submitted without the "_FS_" portion. This caused the config to not reflect properly in the kernel which broke the software installation and other services like systemWebServer that depends on the configs.

This commit amends the relevant EXT4 configs.

@ni/rtos 

Signed-off-by: wkoe <wilson.koe@ni.com>


gratian (changes requested):
Please update the commit message to follow upstream standards for a commit that fixes another commit. See the "Describe your changes" section in  https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/process/submitting-patches.rst?h=v4.14.212

In particular saying this "The previous commit was submitted without the "_FS_" portion" is not helpful because commits in git can be and are re-ordered. Moreover the "previous commit" is already different from the one that introduced the issue so the message is already wrong/confusing.

Your commit should start with something like:
`"Commit 9e8c398d23b6 ("nati_zynq_defconfig: Enable ext4 driver") contained a typo ...."`

The other more important issue with your commit is the fact that it appears you have not used 'make savedefconfig' to generate the new defconfig. If I do the following sequence of commands after merging your change the resulting diff should be empty (and it's not):

```
$ make mrproper
$ make nati_zynq_defconfig
  HOSTCC  scripts/basic/fixdep
  HOSTCC  scripts/kconfig/conf.o
  SHIPPED scripts/kconfig/zconf.tab.c
  SHIPPED scripts/kconfig/zconf.lex.c
  HOSTCC  scripts/kconfig/zconf.tab.o
  HOSTLD  scripts/kconfig/conf
#
# configuration written to .config
#

$ make savedefconfig
scripts/kconfig/conf  --savedefconfig=defconfig Kconfig

$ mv defconfig arch/arm/configs/nati_zynq_defconfig 

$ git diff
diff --git a/arch/arm/configs/nati_zynq_defconfig b/arch/arm/configs/nati_zynq_defconfig
index 338233e30d40..2ecb9ffe9f3c 100644
--- a/arch/arm/configs/nati_zynq_defconfig
+++ b/arch/arm/configs/nati_zynq_defconfig
@@ -363,7 +363,6 @@ CONFIG_EXT2_FS=y
 CONFIG_EXT2_FS_XATTR=y
 CONFIG_EXT2_FS_SECURITY=y
 CONFIG_EXT3_FS=y
-CONFIG_EXT4_FS=y
 CONFIG_EXT4_FS_POSIX_ACL=y
 CONFIG_EXT4_FS_SECURITY=y
 # CONFIG_DNOTIFY is not set
```

NIwkoe:
@gratian
Thanks for your feedback. I've amended the commit message as per suggested. Also, after tinkering with make nati_zynq_defconfig, make menuconfig and make savedefconfig, I finally understood some of the principles of configs. I also realised that EXT3 is no longer needed as it is already replaced by EXT4, which is backward compatible. Hence, I'm removing it in this commit.

gratian (approved):
Reworded to add 'Fixes' tag line per kernel commit message conventions and submitted as commit 65327cf3fefeb7dff82211ad9bc8a3fbf5535082 ("nati_zynq_defconfig: Fix typo error in commit 9e8c398d23b6")

PR #12, opened 2020-12-18
VirtualBench changes for supporting additional variants

bstreiff:
Multiple changes in this PR in support for additional upcoming variants.
- Reorganize the device trees. Previously we had `ni-lci.itb` (VB-8012), `ni-lci-debug.itb` (debug VB-8012 with Ethernet population option) and `ni-VB8034.dtb` (which, despite names, was used for both VB-8034 and VB-8054). Instead use the power of device-tree includes to pull things "common to all VirtualBench" into "common to the individual VB1/VB2 platforms" into common includes. With that done, we can now easily add device tree variants for versions without wireless, which is now as easy as `status="disabled"`.
- Push the mass-storage and RNDIS interfaces in the `g_lci` gadget driver to being boot-time selectable (via cmdline parameters). Between this and using the device tree to declare the state of Ethernet, this means that we can drop down to a single kernel configuration for all products. `ni_lci_debug_defconfig`, despite the name, was the "VirtualBench With Ethernet" config and was being used on VB-80x4 products, so that is now the one we're using everywhere; this means a slight kernel size increase for VB-8012 (because it now builds in unused Ethernet/PHY drivers) but IMO ease of maintenance wins.


gratian (approved):
Ack'ed. Nice clean-up.

bstreiff:
Added `Acked-By`s and merged manually.

PR #14, opened 2021-01-07
nati_x86_64_defconfig: Disable the cpufreq driver to improve RT performance

gratian:
The default configuration on NI Linux RT is optimized for performance
rather than power savings. As such the ability of the cpufreq driver
to dynamically scale the CPU frequency is not used.

cpufreq adds code in the scheduler hot paths and results in increased
latency for real-time threads. Disable it.

Natinst-AZDO-ID: 59877
Natinst-AZDO-ID: 1214772


amstewart (approved):

gratian:
Rebased on top of v5.10.4-rt22 and pushed as commit 7a62a9a9511dd849d391f6707eb7b2fdd55fcb29 ("nati_x86_64_defconfig: Disable the cpufreq driver to improve RT performance")

PR #15, opened 2021-01-08
spidev: add "spidev" compatible string to surpress warning

NIwkoe:
Linux kernel outputs the following complain upon startup:
"spidev spi0.0: buggy DT: spidev listed directly in DT"
and many more lines of obnoxious warning messages.

In actuality, the device tree should describe the spi device with its
"brand,product" names instead of a generic "spidev". For simplicity's
sake (and maintainability), let's add "spidev" into the compatible list.
I think this is the least intrusive implementation to prevent old (eg.
roboRIO) and new devices that also uses spidev from the warning.

This commit adds "spidev" into the spidev_dt_ids list.

@ni/rtos 

Signed-off-by: wkoe <wilson.koe@ni.com>

### DIFF
--- a/drivers/spi/spidev.c
+++ b/drivers/spi/spidev.c
@@ -726,7 +726,7 @@ static int spidev_probe(struct spi_device *spi)
 	 */
 	if (spi->dev.of_node && !of_match_device(spidev_dt_ids, &spi->dev)) {
 		dev_err(&spi->dev, "buggy DT: spidev listed directly in DT\n");
-		WARN_ON(spi->dev.of_node &&
+		WARN_ON(0 && spi->dev.of_node &&
 			!of_match_device(spidev_dt_ids, &spi->dev));
 	}
 


gratian:
@jhershbe I would appreciate your input on this too.

gratian:
@NIwkoe can you describe in a comment here what fixing it the right way involves? I'm worried that while this is a one line change now it will definitively create conflicts on any future merges/upgrades so I'm not sure doing it this way in the long run it is less intrusive (not to mention it being a hack).

In the meantime I will try to educate myself in this area of the kernel so I can give you a better review (also hoping other people can comment/review).

bstreiff:
There's some backstory here: https://lore.kernel.org/linux-spi/20171129150435.5npybjuu2ndmtx2v@senary/

Upstream's argument is, essentially, that "spidev does not describe a device, therefore it should not be in the device tree". There was at least an attempt to add a sysfs interface such that userspace could instantiate and claim a SPI device, https://lore.kernel.org/linux-spi/20171221200309.17967-1-kyle.roeschley@ni.com/, although that's not in 4.14 and doesn't look like it's made it into any other upstream release.

Note that the kernels for Raspberry Pi and Beaglebone also have commits that work around the warning, because they have a similar use case: they want to support SPI as a userspace interface but don't know what a user is necessarily going to plug into the board headers, and they don't want to force users to tweak a device tree. So "hack around the warning" is at least, not entirely without precedent...
- https://github.com/raspberrypi/linux/commit/8caac6a83b5f681602cbd084e5585b7ff14b1384
- https://github.com/beagleboard/linux/commit/c100f1531bb29cc8ff9539235803f487df36bc82

gratian:
Thanks Brandon, that's very useful info and makes me feel better about this commit. It does make me wonder if we should go the Raspberry PI route and disable the warn instead.

NIwkoe:
Thanks Brandon! I think Brandon's response answered the implementation question of this commit. Should we stick to adding "spidev" as compatible or use the Raspberry PI route of disabling the warn? I'm ok with both since they are technically still hacky workarounds.

gratian:
OK. Let's go with the Raspberry PI method of disabling the WARN.
It is less likely to have other side effects now or in the future.

gratian:
The change looks good, but the commit message needs rewording to match the implementation change.

NIwkoe:
Thanks @gratian, I've updated the commit message.

gratian (approved):

PR #22, opened 2021-02-11
nati_zynq_defconfig: Enable SCTP

bsilver8192:
SCTP is designed for flexible datagram delivery with control over the
queueing order, which makes it well-suited for realtime communications
on a robot. We use it on our robot to communicate with coprocessors.

You can't build this module out of tree because enabling it adds the
sctp member to struct net, which requires rebuilding the whole kernel.
Because it's a module, it doesn't impose any other runtime cost.

Signed-off-by: Brian Silverman <bsilver16384@gmail.com>


gratian (approved):
Thanks. I appreciate the contribution.

PR #30, opened 2021-08-19
nati_x86_64_defconfig: add support for Logitech G920

stephendb:
Adds support for logitech steering wheels for driver-in-the-loop users.

Signed-off-by: Stephen Butler <stephen.butler@ni.com>


gratian:
Thanks. I appreciate the contribution.

PR #32, opened 2021-08-30
i2c-i801.c: Skip SPD initialization on cRIO-903x

billpittman:
The cRIO-903x architecture is sort of special, we do not connect
a dimm to the SPD because we use flash instead, but the SPD is still
present because the processor expects it.  However, enumerating and
registering the SPD i2c bus is causing interrupt storms, so for now
we skip the registration on all 903x devices.

Fixes: 01590f361e94 ("i2c: i801: Instantiate SPD EEPROMs automatically")
Signed-off-by: Bill Pittman <bill.pittman@ni.com>
Natinst-AZDO-ID: 1573148

### DIFF
--- a/drivers/i2c/busses/i2c-i801.c
+++ b/drivers/i2c/busses/i2c-i801.c
@@ -1311,6 +1311,8 @@ static void register_dell_lis3lv02d_i2c_device(struct i801_priv *priv)
 /* Register optional slaves */
 static void i801_probe_optional_slaves(struct i801_priv *priv)
 {
+	const char *product;
+
 	/* Only register slaves on main SMBus channel */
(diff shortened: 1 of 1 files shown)


billpittman:
@gratian

bstreiff:
Would it be better to identify 903x-type devices instead of doing this revert? This function is already doing machine identification (`if (dmi_name_in_vendors("FUJITSU")`) so adding a new machine-specific check here doesn't seem too out-of-place.

billpittman:
Good idea @bstreiff.

gratian (approved):

bstreiff (approved):

PR #57, opened 2022-03-15
nati_x86_64_defconfig: Build Hyper-V support in

chaitu236:
Build required Hyper-V modules into kernel as we no longer have
initramfs in runmode to load them.

### Testing
Built kernel and modules locally, copied them over to rootfs partition in a HyperV target and verified runmode boots.

### Note
Hyper-V keyboard module is still built as a module since it depends on SERIO which is configured as a module. Leaving it as is since it is not required for loading rootfs.


amstewart (approved):

gratian (approved):
LGTM. For posterity this adds ~78 kB to the 5.10 bzImage which is an acceptable tradeoff.

gratian:
Pulled. Thanks.

PR #67, opened 2022-06-27
nati_x86_64_defconfig: readd CONFIG_SERIAL_8250_RUNTIME_UARTS=0

bstreiff:
When we rebased the defconfig options from 4.14 to 5.6 (and then further to 5.10), the option CONFIG_SERIAL_8250_RUNTIME_UARTS=0 was accidentally dropped.

Without this, with the default option of 4, the 8250 driver will automatically add four hardcoded serial ports at the legacy PC serial port addresses (from the SERIAL_PORT_DFNS array), regardless of whetheror not they're present in the ACPI tables. On sbRIO platforms, only the first serial port is at 0x3F8-- the remaining ones are at different locations, so with the default setting we get a ttyS1-ttyS3 that are at invalid addresses and shift the remaining eight serial ports (which do come from ACPI) down to new indices, potentially breaking userspace.

This commit restores the behavior as shipped with 4.14.


amstewart (approved):

amstewart:
@bstreiff Can we get this applied to the `5.10` mainline as well? It might be a minute before we switch the NILRT kernel recipes to use 5.15 and I don't want to accidentally lose this change if we decide not to go with 5.15, for whatever reason.

gratian (approved):

bstreiff:
> Can we get this applied to the 5.10 mainline as well? 

Can do, I'll make a PR for that too.

PR #75, opened 2022-08-26
Revert some upstream commits to get USB gadget working

chaitu236:
Reverting two upstream commits because the assumption that
any baytrail hardware without ACPI entries for USB PHY "reset" and
"cs" will have GPIO controller INT33FC in their ACPI tables - which
is not the case for some of our targets.

### Testing
With these reverts, `usb0` interface appears in cRIO-9030, cRIO-9053, cRIO-9037, cRIO-9043, cRIO-9049.


gratian (approved):

chaitu236:
@gratian / @bstreiff can this be merged if there are no reservations.

PR #76, opened 2022-08-26
e1000e: Replace PREEMPT_RT_FULL with PREEMPT_RT

shruthi-ravi:
Commit a555b1a added a delay after long writes on PREEMPT_RT-enabled
kernels, but using the PREEMPT_RT_FULL option, which has since been
replaced by just PREEMPT_RT.
Therefore, 
 - update the delay condition to check on PREEMPT_RT instead.
 - readd CONFIG_E1000_DELAY to nati_x86_64_defconfig

AzDO WI: [2114543](https://ni.visualstudio.com/DevCentral/_workitems/edit/2114543)


gratian:
@shruthi-ravi it looks like CONFIG_E1000_DELAY got dropped from `nati_x86_64_defconfig` probably when it was regenerated and because of the PREEMPT_RT_FULL getting renamed. Can you add a commit to re-enable it in defconfig?

Also, can you submit a PR against the 5.15 branch with these changes too since it will be the new default kernel? I will pull this in 5.10 for completion since it affects it.

amstewart (approved):

gratian (approved):

PR #99, opened 2023-01-18
nati_x86_64_defconfig: enable P2PDMA and its dependencies

alynch-ni:
This feature is needed to support P2P communication between FPGAs and RDMA NICs or NVMe drives.

Signed-off-by: Andrew Lynch <andrew.lynch@ni.com>

I validated that the kernel builds successfully and loads on a PXIe-8880 controller in the place of the shipping kernel in the 2022Q4 image.


gratian (approved):
Looks good, thanks.
For future reference this adds ~40KBytes to the kernel image (bzImage).

PR #100, opened 2023-01-20
serial: 8250_ni: new driver for NI UARTs

bstreiff:
The National Instruments (NI) 16550 is a standard 16550 with larger FIFOs and embedded RS-232/RS-485 transceiver control circuitry. This patch adds a driver that can operate this UART.

---

This patch set is a combination and reformulation of
- 2d4c94552f56 serial_core: Provide a transceiver interface
- 8e9e45935119 8250: Allow client drivers to control transceivers
- a07f2496f126 8250: Add a driver for NI 16550 UART
- 6fbc61e57e12 8250_pnp: Make NIC7750 ID spawn a NI 16550 port
- 8d2f96d747ef2 8250_pnp: Support dual-mode UART through PMR
- 053fe96f7114 8250_pnp: NI ACPI IDs start with NIC
- 7e4f38dcf35eb 8250: NI 16550 can have different FIFO sizes
- 806fb0d47a15 8250_pnp: Provide NI_16BYTE_FIFO flag
- 46312fa9d499 8250_pnp: Make NIC7772 spawn a dual-mode UART
- 3089f71dadd4 8250_pnp: Add new ID NIC792B for NI 16550 UART 25MHz
- 266b84da2486 serial: 8250_ni16550: Add support for NIC7A69
- bb3aac8e6975 8250: Use proper set_mctrl callback for ni16550 clock divider
- d0e236dcb51b 8250_ni16550: fix compiler warning about mixed declarations and code
- 145e8921120b 8250_ni16550: pass termios as param to ni16550_config_rs485

There are significant differences between that pile of patches and what we have here, which is:
- The operative configuration token is `CONFIG_SERIAL_8250_NI` instead of `CONFIG_SERIAL_8250_NI16550`-- this lets the two implementations live side-by-side in our tree (rationale below)
- The generic "transceiver interface" that was added in the earliest patches is now gone-- transceiver enable/disable just happens in the port startup/shutdown callbacks now. Having that as a separate API might have been nice, but I think it'd only be justifiable if I ported other 8250 drivers to use it, and the things that it did just happened in startup/shutdown anyway so it's not really a "missing" function of the 8250 core.
- We're now our own `platform_device` instead of a pile of custom initialization bodged into the `8250_pnp` code. This pushes more of the NI16550-specific code into its own file (and reducing future merge conflicts)
- Because we're now our own `platform_device`, on x64, device enumeration happens a little bit differently than enumeration via the legacy PNP system. This should be invisible to most users unless they go digging real hard in sysfs (`$(realpath /sys/class/tty/ttyS1)` on a cRIO-904x becomes `/sys/devices/pci0000:00/0000:00:1f.0/NIC792B:00/tty/ttyS1` instead of `/sys/devices/pnp0/00:01/tty/ttyS1`-- however most users are probably just going to be using `/dev/ttyS1`, which doesn't change.)
- `PORT_NI16550_*` is gone; we just use `PORT_16550A`. This is a visible change to userspace (`/sys/class/ttyS1/type` and also via `serial_struct::type` with the `TIOCGSERIAL` ioctl), but userspace could not rely on the values we had been using anyway (since we had to keep renumbering on every kernel upgrade). Note that use of this field for identification purposes seems to have fallen out of fashion-- in fact for USB serial devices as of f64d74a59c47 they all just return `PORT_UNKNOWN`.
- `NI_16BYTE_FIFO` is gone now (as are the distinct `ni,ni16550-f16` and `ni,ni16550-f128` OF names)-- instead we look at the `TFS`/`RFS` registers; these have been present in the common UART core since at least 2012, and if they're somehow not there, then 128 bytes seems fine as an assumption, since the single 16-byte-FIFO instantiation seems to have been the exception.
- `NI_CAP_PMR` has a slightly more interesting tale-- it was added to the common core in 2014, shipped in the CVS product line, and then _removed_ with no explanation in 2015. (Filed AzDO 2275198 to HW about that.) At any rate, in the current version of the register map, that register is now no longer reserved and there's risk that it could be reused for some future purpose, so we have to keep the magic flag. :(
- I tried to make the "how do we figure out the UART clock and/prescaler/transceiver" codepath for all the different ACPI ids easier to follow. There are some slight behavioral differences here-- ACPI device properties can technically now override clock frequency and prescaler with any of the ACPI IDs (and not just NIC7A69), but the earlier IDs shouldn't have any defined, so there should be no difference in practice.
- There is documentation for the `ni,ni16550` OF ID because `checkpatch.pl` complains about it. `checkpatch.pl` also complains if you don't put it in its own commit, which is why it's separated. I didn't actually try to give this a spin on a device using OF-based enumeration (aka Zynq-based cRIO) due to the effort involved in getting this version of the driver working with the current Zynq kernel.
- Replaced the boilerplate license text with the SPDX identifier for that same text.

This is hopefully a lot cleaner and more upstream-acceptable-- but because it's a significant reworking, I'd like for it to simmer in our trees for a little bit to get some ATS runtime (in particular, get some testing on more than just a cRIO-904x with hand-testing with pyserial!). If it causes problems, we have a quick escape hatch (we can flip the kconfig token back)-- if it turns out good, then we can start tackling reverts on the older implementation.

You might also notice that "8250_ni" is named kind of generically-- I did consider trying to add support for PCI/PXI devices, at least to the extent of the since-reverted attempt in fdc2de87124f ("serial/8250: Add support for NI-Serial PXI/PXIe+485 devices."). I think this is viable to add to 8250_ni (the RS-232 control register looks the same, so this mostly amounts to "add PCI enumeration and `num_ports`/address offsetting)), but wouldn't be a total replacement for the NI-Serial ni843x driver (lack of DMA). Maybe later...

I went with citing Jaeden and Karthik as the original authors-- most of the other changes since then have been mostly-trivial adding of IDs and/or handling the effects of 8250 core changes. Let me know if you think that should change.


bstreiff:
@ni/rtos

bstreiff:
also pinging @sjasonsmith who also expressed a desire to look at this but whom I cannot seem to add as a reviewer

drivers/tty/serial/8250/8250_ni.c
```diff
@@ -0,0 +1,461 @@
...
+	if (port->rs485_config)
+		ret = ni16550_enable_transceivers(port);
+
+	return ret;
+}
+
+static void ni16550_port_shutdown(struct uart_port *port)
+{
+	if (port->rs485_config)
+		ni16550_disable_transceivers(port);
```
sjasonsmith: It looks like the prior implementation was calling enable/disable for all ports, not only RS-485. I haven't looked at any schematics to confirm whether transceiver enable pins are wired up on the RS-232 implementations where this is instantiated, but they certainly were on PCI implementations which share this register interface.

sjasonsmith: I tried poking around some schematics and it looks like at least some RS-232 transmitters were not wired to require this, but I think it's still best to always call enable/disable. This has been used in enough different designs that I wouldn't want to change this behavior without reviewing all of them.

sjasonsmith:
> The generic "transceiver interface" that was added in the earliest patches is now gone-- transceiver enable/disable just happens in the port startup/shutdown callbacks now. Having that as a separate API might have been nice, but I think it'd only be justifiable if I ported other 8250 drivers to use it

I think that when this was originally created there were not yet upstream mechanisms for RS-485 configuration. It's role decreased as it was merged with newer kernel versions, until finally reaching complete obsolescence in this PR.

drivers/tty/serial/8250/8250_ni.c
sjasonsmith: I'm almost positive that PCI boards were wired to require the enable_transceivers call for RS-232, although that is not particularly relevant right now. You can actually see that the 8250_pci driver set the transceiver enable bit for 8430 (RS-232) support:

https://github.com/ni/linux/blob/7a39588dec6662a33e1bd77d4347687d4adbd372/drivers/tty/serial/8250/8250_pci.c#L772

drivers/tty/serial/8250/8250_ni.c
```diff
@@ -0,0 +1,461 @@
...
+				dev_vdbg(port->dev, "2-wire Auto\n");
+				pcr |= NI16550_PCR_AUTO_RS485;
+			} else {
+				/* DTR-controlled */
+				/* No Echo */
+				dev_vdbg(port->dev, "2-wire DTR no echo\n");
+				pcr |= NI16550_PCR_DTR_RS485;
+			}
+		}
+	} else {
```
sjasonsmith: Recent upstream changes in 63f75fea3a72f93a022d2f027842fe56dd0ebb08 have revealed some misalignment between the NI interpretation of these and what they are meaning for the rest of the kernel. Some of that probably stems from our original implementation pre-dating the serial_rs485 structure.

The way the Linux kernel has currently defined RS-485 support, only "4-Wire / RS-422" and "2-Wire Auto" really make sense. Setting `SER_RS485_ENABLED` implies that you want the driver to handle enabling/disabling the transceiver for you, and the other flags only alter the methods used for automatic control.

We had interpreted this differently, with `SER_RS485_ENABLED` meaning that you were half duplex, but allowed for both automatic and manual control modes.

I think this should be reworked to align with the upstream interpretation. This means eliminating the "2-wire DTR no echo" and "2-wire DTR with echo" implementations. Those modes have always been hidden in NI-VISA on NI Linux RT, so I don't anticipate there are many users of those modes, if any at all.

I have some other documentation written up as we've looked at this for other drivers. Reach out to me for more details.

sjasonsmith: I thought the ni16550 UART would require writing to the ACR register to enable automatic transceiver control, in addition to the PCR. I don't see that here, nor do I see it in the 8250_port driver. This is something to dive into more to understand how this is (or isn't) working.

drivers/tty/serial/8250/8250_ni.c
bstreiff: We were only calling enable/disable_transceivers on RS-485 ports before:
- port->txvr_ops was set in `ni16550_port_setup`: https://github.com/ni/linux/blob/nilrt/master/5.15/drivers/tty/serial/8250/8250_ni16550.c#L203
- and we only ever called `ni16550_port_setup` for RS-485 ports, not RS-232 ones: https://github.com/ni/linux/blob/nilrt/master/5.15/drivers/tty/serial/8250/8250_pnp.c#L570

The only RS-232 ports that would be using this driver (or `8250_ni16550`) would be the dual-mode ports on the CVS line and any RS-232 ports on sbRIOs. The RS-232 ports on other devices would be using the same UART hardware IP core, but their ACPI tables use the `PNP0501` identifier and are thus loaded by `8250_pnp` which doesn't know about the `PCR` register, so those devices wouldn't have been able to expect the transceiver-enable bit be toggled.

drivers/tty/serial/8250/8250_ni.c
bstreiff: The old implementation never did anything with `ACR` before  (https://github.com/ni/linux/blob/nilrt/master/5.15/drivers/tty/serial/8250/8250_ni16550.c#L9) but I did notice that when we tried with the PCI/PXI ones in fdc2de87124f we do write to `ACR` there.

sjasonsmith: > The old implementation never did anything with ACR

OK, looking through some of the implementation as stirred up some old memories. Prior to existing in the NI16550 form, the PCR was implemented in external logic. It was necessary to program both the ACR (for automatic DTR control) and PCR (to configure external logic) to get the proper automatic transceiver control.

In the NI16550 it is all integrated, so there are explicit RxEnable and TxEnable lines/signals. This allows for automatic transceiver control without explicitly configuring the UART's Auto-DTR through the ACR register.

So, not programming the ACR should be fine where this is used. If it is ever applied to PCI-based boards then more work will be needed to configure the Auto-DTR behavior.

drivers/tty/serial/8250/8250_ni.c
sjasonsmith: > and we only ever called ni16550_port_setup for RS-485 ports, not RS-232 ones: https://github.com/ni/linux/blob/nilrt/master/5.15/drivers/tty/serial/8250/8250_pnp.c#L570

That detail escaped me, so I agree you are fine.

gratian (approved):
Pending @sjasonsmith approval before merging.

drivers/tty/serial/8250/8250_ni.c
sjasonsmith: I think this is fine for now, with the understanding that we'll need to revisit the RS-485 mode configuration.

sjasonsmith (approved):
Approve from my perspective. I can't verify that the details of all the different ACPI IDs is correct.

Wire modes will need to be revisited later.

drivers/tty/serial/8250/8250_ni.c
bstreiff: While I was reducing the cases down, I also I went looking at adding ACR programming (since, skimming through the HDL, it _looks_ like it should be harmless to do on these instantiations and makes us ready for the PCI ones?)... and I think I figured out why fdc2de87124f didn't work (acr |= NI16550_ACR_DTR_MANUAL_DTR doesn't work when that constant is zero, so once you acr |= NI16550_ACR_DTR_AUTO_DTR'd you were stuck that way, oops)

bstreiff:
changes in v2:
- clean up comments in header (it's not just a "transceiver driver")
- use `BIT` macro
- rename `ni16550_config_rs485` -> `ni16550_rs485_config` to match struct member it is assigned to
- remove 2-wire DTR with/without echo cases
- don't update `port->rs485` (`uart_set_rs485_config` does this for us)
- use `rs485_supported` to declare support
- program `ACR` for parity with PCI devices
- enable/disable transceivers on all devices (also for parity with PCI)
- mark the SPDX-License-Identifier in dt bindings actually as a SPDX-License-Identifier
- add an extra commit to mark `txvr_ops` as deprecated (note AzDO#2280538)

gratian:
I've reset approvals in order to review v2.

gratian (approved):

include/linux/serial_core.h
```diff
@@ -567,7 +567,7 @@ struct uart_port {
 	unsigned int		frame_time;		/* frame timing in ns */
 	unsigned int		type;			/* port type */
 	const struct uart_ops	*ops;
-	const struct txvr_ops	*txvr_ops;		/* transceiver ops */
+	const struct txvr_ops	*txvr_ops;		/* transceiver ops (deprecated) */
```
sjasonsmith, line 570: I actually think that the way the other driver is configured it would compile/link fine with it completely removed, and the "enable transceiver" it actually performed was redundant with other code anyway. I understand wanting to keep it marked deprecated temporarily while other drivers are cleansed, but I don't think you should have to wait long for the complete removal.

drivers/tty/serial/8250/8250_ni.c
```diff
@@ -0,0 +1,449 @@
...
+						 port);
+
+	/* "rs485" should be given to us non-NULL. */
+	if (WARN_ON(rs485 == NULL))
+		return -EINVAL;
+
+	pcr = serial_in(up, NI16550_PCR_OFFSET);
+	pcr &= ~NI16550_PCR_WIRE_MODE_MASK;
+
+	if (rs485->flags & SER_RS485_ENABLED) {
```
sjasonsmith, line 106: I think it is fine that `SER_RS485_RTS_ON_SEND` and `SER_RS485_RTS_AFTER_SEND` are treated as "don't care", since our hardware doesn't need them to implement automatic transceiver control.

What about `SER_RS485_RX_DURING_TX`? Should this generate an error if a user tries to set that, since we can't honor that request?

sjasonsmith: And I've answered my own question! I see that you've implemented rs485_supported to prevent that.

`SER_RS485_RTS_ON_SEND` has to remain "supported" since all of our existing code uses it. `SER_RS485_RTS_AFTER_SEND` was probably never used by us, but leaving it as a "don't care" doesn't bother me.

drivers/tty/serial/8250/8250_ni.c
```diff
@@ -0,0 +1,449 @@
...
+
+	/* Set prescaler to CPR register. */
+	serial_out(up, UART_SCR, UART_CPR);
+	serial_out(up, UART_ICR, prescaler);
+}
+
+static const struct serial_rs485 ni16550_rs485_supported = {
+	.flags = SER_RS485_ENABLED | SER_RS485_RTS_ON_SEND | SER_RS485_RTS_AFTER_SEND,
+	.delay_rts_before_send = 1,
+	.delay_rts_after_send = 1,
```
sjasonsmith: Is there a reason to say these are supported? Since the transceiver control is handled entirely in hardware, these won't actually impact the timing of when transceivers are turned on/off.

I haven't studied _every_ usage of them in serial_core and 8250_port. It is possible you might get some software delays around your transmits, but without the impact on signal timing the feature implies.

sjasonsmith (changes requested):
I consider my comments pretty minor, and I'm fine with what you decide is the proper reaction to them.

I'm not 100% of GitHub workflows. I'm marking this as "Request Changes" but my intent is that after you read my comments you can decide how to proceed and do so without having to wait on me.

drivers/tty/serial/8250/8250_ni.c
bstreiff: I think we had been "supporting" those with the old implementation (`ni16550_config_rs485` never checks them for being non-zero), and that's what `serial8250_em485_supported` uses by default. It's possible that the results were bogus.

Unfortunately I don't have local hardware on which to throw a scope. :/

There do seem to be serial ports in here that force them to be = 0 (or, more accurately, leave them unset). Removing them might be the right call, especially since it doesn't look like any of our code ever touches that option.

include/linux/serial_core.h
bstreiff: Yeah, due to the way this feature was used, I'm not sure that this one needs the whole "we can't get rid of this for three years until a NI-Serial that needs it is no longer supported" plan, but it seems reasonable to mark it this way until we've confirmed that.

bstreiff:
v3 drops the rts delay stuff out of the rs485-supported struct.

sjasonsmith (approved):

bstreiff:
Merged into `nilrt/master/6.0`

PR #106, opened 2023-02-28
nati_x86_64_defconfig: Add support for additional storage features

alynch-ni:
This change enables the use of remote NVMe-oF storage over RDMA. It also enables the creation of MD software RAID arrays for compatibility with arrays used on other systems.

I built the kernel locally with this change and tested it on a PXIe-8880 controller. I verified that I can mount NVMe-oF arrays connected to an RDMA capable NIC and that I can use the same software RAID arrays from a desktop Linux distribution.


gratian:
@alynch-ni Can you regenerate the defconfig using `make savedefconfig`? That collapses the options that can be deduced from other config options (and makes future updates to defconfig cleaner).

The steps would be:

- `make nati_x86_64_defconfig`
- `make menuconfig` and enable the options needed
- `make savedefconfig`
- `mv defconfig arch/x86/configs/nati_x86_64_defconfig`
- update commit with new defconfig

If you are starting from the already modified defconfig in this PR you can skip the `make menuconfig` step.

When you run it through `make savedefconfig` you will notice that:
`CONFIG_NVME_FABRICS=m` doesn't show up in the diff anymore because it is implied by the other options.

gratian:
@alynch-ni Can you also add a comment here with what is the size delta of /lib/modules after adding these (for future reference)?

alynch-ni:
I amended the commit to have the file as generated by `make savedefconfig`.
The /lib/modules size increased by 1.1MB from this change.

gratian (approved):

PR #108, opened 2023-03-08
Create a method to increase the max number of PPS sources

jeminor:
For a fully loaded PXI chassis, a system can have more than 16 PPS sources.  Make that configurable and set the default to 2x what it is today.

Testing:
TBD, will post back once this has been validated.


include/uapi/linux/pps.h
```diff
@@ -26,7 +26,6 @@
 #include <linux/types.h>
 
 #define PPS_VERSION		"5.3.6"
-#define PPS_MAX_SOURCES		16		/* should be enough... */
```
bstreiff, line 29: I fear that removing this constant might fun afoul of the "we don't break userspace" rule, but I don't see any other attempts elsewhere to plumb a kernel-config-defined constant into the uapi headers in order to maintain it accurately.

At the same time, uapi-header-defined constant limits are kind of bad in a world where we have sysfs and something can just enumerate /sys/class/pps, and at least in a github code search I see no other uses of that constant that are not coming from the many copies of the Linux kernel tree.

I'm okay with this in our tree but I know it's likely to become an issue when it goes upstream...

bstreiff (approved):

include/uapi/linux/pps.h
jeminor: Any suggestions here on how to avoid that?  I might continue to refine before sending upstream, but may also make that a separate change...

bstreiff: The simplest might be to just change the comment from `/* should be enough... */` to `/* deprecated, use sysfs instead */`.

gratian (approved):

include/uapi/linux/pps.h
jeminor: fixed in v2.

bstreiff:
LGTM, just waiting on the testing feedback before I pull it.

jeminor:
Looks like the clients here want to make more like 128 of these in some instances, so I think making this run/load time configurable is a better solution.  Will refactor or let someone else do that.

SparkingSpork:
Closed, see #118

PR #109, opened 2023-04-06
Merge tag 'v5.15.106' into nilrt/master/5.15

gratian:
This is a regular stable merge of latest linux-stable tag `v5.15.106`.

No merge conflicts encountered.

#### Testing

- [x] boot tested on cRIO-9042, and PXIe-8840.
- [x] diffed dmesg against previous version; no new errors.
- [x] re-versioned NI drivers on cRIO-9042 via `dkms autoinstall` and verified there are no errors.

Additional functional and performance testing will be handle by ATSs.


chaitu236:
> This is a regular stable merge of latest linux-stable tag v5.15.106.

@gratian  We're merging 5.15 `linux-stable` here (instead of `linux-stable-rt` like usual) - does this imply that `linux-stable` contains all the rt patches we care about?

gratian:
@chaitu236 If you are asking if the `linux-stable` tree has now all the rt patches the answer is no. The rt patches still come from our tree which is based on `linux-stable-rt`. This just merges the latest fixes in `linux-stable` into our tree (which already has the rt patches we need).

FWIW It is the same process the `linux-stable-rt` tree follows i.e. merges form `linux-stable` with occasional fixes for conflicts with rt patches. In this case there were no conflicts so it's safe to merge directly from stable and not wait for the stable-rt tree to catch up.

drivers/media/usb/uvc/uvc_video.c
```diff
@@ -1951,6 +1953,17 @@ static int uvc_video_start_transfer(struct uvc_streaming *stream,
 			"Selecting alternate setting %u (%u B/frame bandwidth)\n",
 			altsetting, best_psize);
 
+		/*
+		 * Some devices, namely the Logitech C910 and B910, are unable
```
bstreiff, line 1957: huh, I use a C910, I should keep this in mind.

bstreiff (approved):
LGTM

PR #110, opened 2023-04-18
activate the cuse kernel modul

UWkerb-NI:
We need the cuse kernel module for our userspace drivers. Since this kernel module has dependencies on the fuse kernel module and its compilation parameters, the cuse kernel module must necessarily be built together in the context of the kernel build.

The patch simply enables the cuse kernel module to build.


bstreiff (changes requested):
Hi @UWkerb-NI

For this repo we try to keep commits in the format prescribed by [the upstream Linux project](https://docs.kernel.org/process/submitting-patches.html). Right now your commit description is simply the line `activate the cuse kernel modul`. Your commit description should:
- give a summary phrase indicating the subsystem to which this patch is targeted
- a description of what the patch is doing
- have a `Signed-off-by` with your name and e-mail address

Obviously since this is a "just add this module to the defconfig" patch, the description will be brief. Commit 4ed5022e29e5 is a good example of form and level of detail that would be expected here.

UWkerb-NI:
the pull request was cancelled.

PR #111, opened 2023-04-18
nati_x86_64_defconfig: enable the cuse module

UWkerb-NI:
Enable the cuse module to support the character device interface in userspace drivers.


gratian:
It looks like `arch/x86/configs/nati_x86_64_defconfig` was edited manually or not passed through `make savedefconfig`. I'm seeing a diff if I use savedefconfig.

Can you use these steps to make the edit:

- start from a clean state `make mrproper`
- load the previous configuration `make nati_x86_64_defconfig`
- make the desired change to enable CONFIG_CUSE for example via `make menuconfig` or other means
- save your changes with `make savedefconfig`; this creates a defconfig file which is the compressed .config
- update nati_x86_64_defconfig with this new change by ` mv defconfig arch/x86/configs/nati_x86_64_defconfig`

This ensures that future edits to `nati_x86_64_defconfig` have clean diffs.

UWkerb-NI:
Hi @gratian i edit the `arch/x86/configs/nati_x86_64_defconfig` manually to keep the changes small.
If you prefer to use the menuconfig workflow, I can do that too.

UWkerb-NI:
Hi @gratian I followed your instructions and pushed the changes.

gratian (approved):
The changes themselves look OK. A couple of comments for changes like this in the future:

- We try to keep the history in the main branches clean so a single commit would be preferable here i.e. you could have forced push an amended commit in your branch that includes the safedefconfig generated change. Not a big deal this time around @bstreiff can squash the two commits into one before pushing to nilrt/master/5.15
- We prefer to place temporary repos/branches into our own accounts space and not under github.com/ni where it could confuse customers in thinking it's some official repo. Once this change gets pulled we will probably want to delete the repo you created under github.com/ni to clean things up.

bstreiff (approved):
Yup, I can squash these down to one commit and then cherry-pick it into the future branches (and then we can clean up that errant fork)

bstreiff:
[squashed and merged into `nilrt/master/5.15` via command-line]

bstreiff:
[also cherry-picked into `nilrt/master/6.0` and `nilrt/master/6.1`]

UWkerb-NI:
Hi @gratian thanks for the correction. I didn't realise that the branch was created under the ni account. I don't know how I could have done that. I thought the branch was always created under the own login. Next time I'll have to look more closely.

I can understand about the clean git history. We do it that way here too. I'm probably spoiled by gitlab. With gitlab you can set it up so that a squash of changes happens automatically. I will combine all the adjustments into one commit next time.

Thank you very much.

PR #118, opened 2023-06-01
pps: Use sysfs module parameter to replace PPS_MAX_SOURCES.

SparkingSpork:
Currently, the PPS_MAX_SOURCES is a hard coded constant. On some systems, the current maximum of 16 is too low. This change converts that logic to use a sysfs parameter instead of a hard-coded value and increases the default value to 32.

## Testing:
- Built and installed new kernel on target. 
   - Confirmed that new `/sysfs/module/pps_core/paramters/pps_max_sources` existed and was set to default of 32. 
   - Confirmed value could be changed.
   - Confirmed invalid values errored.
   - Confirmed that adding `pps_core.pps_max_sources=64` to the boot parameters resulted in a value of 64.
- Ran `checkpatch.pl`

### DIFF
--- a/drivers/pps/sysfs.c
+++ b/drivers/pps/sysfs.c
@@ -8,6 +8,7 @@
 
 #include <linux/device.h>
 #include <linux/module.h>
+#include <linux/moduleparam.h>
 #include <linux/string.h>
 #include <linux/pps_kernel.h>
 
@@ -97,3 +98,12 @@ const struct attribute_group *pps_groups[] = {
 	&pps_group,
 	NULL,
 };
+
+/*
+ * Module parameters
+ */
+
+unsigned int pps_max_sources = 16;
+
+MODULE_PARM_DESC(pps_max_sources, "Maximum number of pps sources that can be allocated.");
+module_param(pps_max_sources, uint, 0664);
--- a/include/linux/pps_kernel.h
+++ b/include/linux/pps_kernel.h
@@ -67,6 +67,7 @@ struct pps_device {
  */
 
 extern const struct attribute_group *pps_groups[];
+extern unsigned int pps_max_sources;
 
 /*
  * Internal functions.
--- a/include/uapi/linux/pps.h
+++ b/include/uapi/linux/pps.h
@@ -26,7 +26,7 @@
 #include <linux/types.h>
 
 #define PPS_VERSION		"5.3.6"
-#define PPS_MAX_SOURCES		16		/* should be enough... */
+#define PPS_MAX_SOURCES		16		/* deprecated, use sysfs instead */
 
 /* Implementation note: the logical states ``assert'' and ``clear''
  * are implemented in terms of the chip register, i.e. ``assert''
--- a/drivers/pps/pps.c
+++ b/drivers/pps/pps.c
@@ -354,7 +354,7 @@ int pps_register_cdev(struct pps_device *pps)
 	 * Get new ID for the new PPS source.  After idr_alloc() calling
 	 * the new source will be freely available into the kernel.
 	 */
-	err = idr_alloc(&pps_idr, pps, 0, PPS_MAX_SOURCES, GFP_KERNEL);
+	err = idr_alloc(&pps_idr, pps, 0, pps_max_sources, GFP_KERNEL);
 	if (err < 0) {
 		if (err == -ENOSPC) {
 			pr_err("%s: too many PPS sources in the system\n",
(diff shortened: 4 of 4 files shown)


include/linux/pps_kernel.h
```diff
@@ -67,6 +67,7 @@ struct pps_device {
  */
 
 extern const struct attribute_group *pps_groups[];
+extern int pps_max_sources;
```
SparkingSpork: Thoughts on `max_sources` instead of `pps_max_sources` since it's a sysfs parameter now?

include/uapi/linux/pps.h
```diff
@@ -26,7 +26,7 @@
 #include <linux/types.h>
 
 #define PPS_VERSION		"5.3.6"
-#define PPS_MAX_SOURCES		16		/* should be enough... */
+#define PPS_MAX_SOURCES		16		/* deprecated, use sysfs instead */
```
SparkingSpork, line 29: Is a `#pragma deprecated(PPS_MAX_SOURCES)` valuable here?

bstreiff: I don't see use of that `#pragma` anywhere else in the uapi headers, and [Linus seems against a `__deprecated` tag](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=771c035372a0). I think annotating it with a comment is the correct thing to do.

drivers/pps/sysfs.c
```diff
@@ -97,3 +98,28 @@ const struct attribute_group *pps_groups[] = {
 	&pps_group,
 	NULL,
 };
+
+/*
+ * Module parameters
+ */
+
+int pps_max_sources = 32;
```
bstreiff: This should probably default to the previous limit of 16; we probably shouldn't change the defaults out from other users.

drivers/pps/sysfs.c
```diff
@@ -97,3 +98,28 @@ const struct attribute_group *pps_groups[] = {
...
+
+static int pps_max_sources_write(const char *val, const struct kernel_param *kp)
```
bstreiff: why not use param_set_int here?

Or for that matter, make the parameter a uint and use param_get_uint/param_set_uint, since it's not meaningful to have a negative number of pps sources.

bstreiff (changes requested):

drivers/pps/sysfs.c
```diff
@@ -97,3 +98,28 @@ const struct attribute_group *pps_groups[] = {
...
+
+	return param_set_int(val, kp);
+}
+
+static const struct kernel_param_ops pps_max_sources_ops = {
+	.set = pps_max_sources_write,
+	.get = param_get_int
+};
+
+module_param_cb(pps_max_sources, &pps_max_sources_ops, &pps_max_sources, 0664);
```
chaitu236: Adding `MODULE_PARM_DESC` may be useful.

include/uapi/linux/pps.h
SparkingSpork: Cool. Dropping this comment then.

drivers/pps/sysfs.c
SparkingSpork: When I looked at the function prototype for one of the places it was used, the function input was an `int`. Looking at the other 2 places though, that's not consistent and the other two accept `unsigned`. I'll switch to `uint`.

SparkingSpork: That being said, I'm also realizing this is used to allocate a character device at module init. So I'll still need a function to handle reallocating when it changes which is quite a bit more complicated.

include/linux/pps_kernel.h
chaitu236: Can you clarify what you mean by `since it's a sysfs parameter now`?
Since this is a global variable, I prefer `pps_max_sources` for more specificity.

drivers/pps/sysfs.c
SparkingSpork: Switched to use `module_param()` with a type of uint since the callback isn't needed.

include/linux/pps_kernel.h
SparkingSpork: My thought was that it will only be referred to with the context of pps. Previously, code might have `PPS_MAX_SOURCES` in it with no other context. Making it a module parameter means that `pps_core` will typically be part of the calling convention indicating it's part of pps. That being said, I think the more specificity is good since it's used in pps without context so I'll leave it as-is.

SparkingSpork:
## V2
- Value is now an unsigned integer.
- Switched to `module_param` instead of `module_param_cb`
- Added `MODULE_PARM_DESC`
- Kept old default maximum of 16

jeminor (approved):
I will direct message you with a validation request.

chaitu236 (approved):

bstreiff:
Closing in favor of the solution in #131

PR #124, opened 2023-07-18
Revert "Revert "mmc: core: enable CMD19 tuning for DDR50 mode""

chaitu236:
This reverts commit 43ccc9ea05d44f59fc231f5fc16acebb07017be7 which is no longer necessary as CMD19 tuning works for DDR50 mode in 6.1 kernels.

WI: [2113488](https://dev.azure.com/ni/DevCentral/_workitems/edit/2113488)

### Testing
- [x] With swissbit 16G micro sd card oon a cRIO-9053, no "ddr50 tuning failed" log in dmesg
- [x] Added logs to mmc code to verify that tuning is happening and not being skipped.
- [x] Following log in dmesg confirming DDR50
         `mmc0: new ultra high speed DDR50 SDHC card at address 0001`
- [x] Read speeds from the micro sd card in 43MB/s range which indicate DDR50 mode (and not high speed mode which is limited to 35MB/s)
- [x] Reproduced issue on the target on older kernel before the mmc patches were put in.

### DIFF
--- a/drivers/mmc/core/sd.c
+++ b/drivers/mmc/core/sd.c
@@ -665,9 +665,25 @@ static int mmc_sd_init_uhs_card(struct mmc_card *card)
 	 * SDR104 mode SD-cards. Note that tuning is mandatory for SDR104.
 	 */
 	if (!mmc_host_is_spi(card->host) &&
-	    (card->sd_bus_speed == UHS_SDR50_BUS_SPEED ||
-	     card->sd_bus_speed == UHS_SDR104_BUS_SPEED))
+		(card->host->ios.timing == MMC_TIMING_UHS_SDR50 ||
+		 card->host->ios.timing == MMC_TIMING_UHS_DDR50 ||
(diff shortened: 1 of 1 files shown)


gratian (approved):
Looks good to me. It would be helpful to get a thumbs up from Kyle Roeschley too.

bstreiff (approved):

kroeschl:
~Did we verify that this specific Swissbit card reproduced the tuning failure on an old kernel without this patch? Just in case this was a later revision of the card or something like that.~

Oops, that's in the PR description. I would suggest that we test with an old, un-patched kernel in case this was a later hardware rev of the SD card.

chaitu236:
> I would suggest that we test with an old, un-patched kernel in case this was a later hardware rev of the SD card.

I tested with an unpatched 4.9 kernel and could reproduce the issue.

bstreiff:
Merged into `nilrt/master/6.1` as fa76b53408dd.

PR #126, opened 2023-08-03
Merge tag 'v6.1.38-rt12' into nilrt/master/6.1

dylan-matthew-turner:
## Justification

This is a regular stable merge of the latest available linux-stable-rt tag v6.1.38-rt12.

[AB#2395763](https://dev.azure.com/ni/DevCentral/_sprints/backlog/RTOS/DevCentral/23C3/08?workitem=2395763)

There were just two conflicts:

### drivers/usb/dwc3/core.h

```
<<<<<<< HEAD (ours)
* @has_dsm_for_softreset: set if we want to use BIOS to do core soft reset
=======
* @suspended: set to track suspend event due to U3/L2.
>>>>>>> v6.1.38-rt12 (theirs)
```

Resolved in place by keeping both. We added has_dsm_for_softreset in (e65d593), and they added suspended in (1ec1452), but neither change is in conflict except these doc comments.

We'd obviously want both sets of doc comments, since after the merge we'll have both struct members.

### mm/maccess.c

```
<<<<<<< HEAD
#include <linux/syscalls.h>
=======
#include <asm/tlb.h>
>>>>>>> v6.1.38-rt12
```

Similar case here. We added a syscall and needed the syscall library (9f4788d), and they changed copy_from_user_nofault and needed the asm/tlb.h library (2e7ad87), but the only conflict happens to be in the includes, so we can simply keep both includes.

## Testing

 - Deployed rebuilt kernel to cRIO-9037
 - Set up and ran dkms autoinstall with the new kernel version
 - Checked dmesg for obvious errors


drivers/usb/dwc3/core.h
```diff
@@ -1110,7 +1110,9 @@ struct dwc3_scratchpad_array {
  *	3	- Reserved
  * @dis_metastability_quirk: set to disable metastability quirk.
  * @dis_split_quirk: set to disable split boundary.
+ * * @has_dsm_for_softreset: set if we want to use BIOS to do core soft reset
  * @has_dsm_for_softreset: set if we want to use BIOS to do core soft reset
```
gratian: Minor nitpick but you've ended up with 2 copies of the comment for "@has_dsm_softreset" after the conflict resolution.

Also it should probably go after the comment for the new "suspend" bit flag that got added (i.e. the comments seem to match the order of the fields in the struct.

gratian (commented):
Looks good - just that minor issue with the conflict resolution for drivers/usb/dwc3/core.h.

dylan-matthew-turner:
Ignore the duplicated commit in the PR feed; it's not in the log. I signed off with the wrong email and had to rebase, amend the message, and re-push

gratian (commented):
Hmm, I was hoping we can fix this without introducing an extra commit we need to worry about when rebasing to a newer version.
At the very least this commit needs a "Fixes" tag which indicates which commit this relates to when we rebase.

Personally I would just redo the merge and do the correct conflict resolution in the first place but I will defer to @bstreiff on if we want to carry this as a separate commit.

Edit:
To clarify I'm talking about commit 46ed65c460d69f19ec52dd2283fe95b3ea220f93 ("drivers/usb/dwc3/core: re-order comments")

dylan-matthew-turner:
II redid the merge.

gratian (approved):
Nitpick: the merge commit ended up with your personal email address for signoff.

dylan-matthew-turner:
Fixed

bstreiff (approved):

PR #127, opened 2023-08-07
Replace "v1" NI 16550 patch with "v5" version

bstreiff:
Drop the originally-submitted-to-upstream version of the NI 16550 patch and replace it with what will go up as "v5".

There is one substantial change in here: the uartclk for NIC792B devices (the ACPI ID used with cRIO-904x) was erroneously transcribed as 25000000 instead of as 22222222 because of the way the `NI_CPR_CLK_25000000` constant had been originally named.

The other changes either relate to device-tree schema correctness or are otherwise formatting and style issues.


gratian (approved):

bstreiff:
applied onto nilrt/master/6.1 as a649b7dd3a88, 734ec768bd09, 1c95ad57bb69, and cf03978daa2b

PR #133, opened 2023-09-08
Revert "usb: dwc3: gadget: Add 1ms delay after end transfer command without IOC"

chaitu236:
This reverts commit 1450c82a16bba232dca0ff667b17c0970ddecb85.

That commit caused usblan to not work reliably between some cRIOs (903x,
9041) and windows host.
The issue happens only sometimes and when it happens, windows host would
report the network adapter as unconnected. The only way to reliably
recover from this was to reconnect usb cable or disable and enable
windows network adapter. Rebooting the cRIO would also sometimes fix the
issue.

Natinst-AzDo-ID: [2419039](https://dev.azure.com/ni/DevCentral/_workitems/edit/2419039)

### Testing
- [x] Verified on cRIO-9034 that USBLAN works with the commit reverted.
- [x] Verified on cRIO-9041 that USBLAN works with the commit reverted.

### Note
I intend to create a tech debt item to fix the original commit (if possible) and upstream it.

### DIFF
--- a/drivers/usb/dwc3/gadget.c
+++ b/drivers/usb/dwc3/gadget.c
@@ -1690,7 +1690,6 @@ static int __dwc3_gadget_get_frame(struct dwc3 *dwc)
  */
 static int __dwc3_stop_active_transfer(struct dwc3_ep *dep, bool force, bool interrupt)
 {
-	struct dwc3 *dwc = dep->dwc;
 	struct dwc3_gadget_ep_cmd_params params;
 	u32 cmd;
 	int ret;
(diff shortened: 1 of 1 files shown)


gratian (approved):
I think the revert is OK but the commit message (for the revert) could use a bit of improvement so we can reason about it in the future.

Can you add a tag with the bug number, something like "Natinst-AzDo-ID" or something like that? And maybe a bit more context in the commit message?

chaitu236:
> I think the revert is OK but the commit message (for the revert) could use a bit of improvement so we can reason about it in the future.
> 
> Can you add a tag with the bug number, something like "Natinst-AzDo-ID" or something like that? And maybe a bit more context in the commit message?

Added some more context and the bug id. Let me know if this is ok.

gratian (approved):

bstreiff (approved):

PR #135, opened 2023-09-14
nati_x86_64_defconfig: enable CONFIG_CGROUP_FAVOR_DYNMODS

gratian:
Per help:

  "This option enables the "favordynmods" mount option by default which
  reduces the latencies of dynamic cgroup modifications such as task
  migrations and controller on/offs at the cost of making hot path
  operations such as forks and exits more expensive."

When this option is disabled (the default) increased latencies have been
observed in some of our real-time benchmarks (ex. FPGA streaming
benchmark). Enable it.


bstreiff (approved):

gratian:
### Testing

- [x] built and booted kernel with modified config on a cRIO-9034
- [x] verified the FPGA streaming benchmark doesn't exhibit the real-time performance regression observed before

### Note to maintainers

- The upstream change adding this config option (and changing the default behavior for RT) was introduced in 6.0
- We need to apply this change to `nilrt/master/6.1` and `nilrt/23.8/6.1`

bstreiff:
pulled into both nilrt/master/6.1 and nilrt/23.8/6.1

PR #138, opened 2023-10-20
Upgrade 'next' development nilrt kernel from 6.1 to 6.6

mike-petersen-ni:
This is a rebase of our commits from `nilrt/master/6.1` to the latest devel-rt branch, `linux-6.6.y-rt`, which is currently tagged as `v6.6-rc6-rt10` (e7eb3321622c).

[AB#2484319](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2484319)

### 9f4788d6985f -> 9c5b3581cf1a

<details><summary>arch/x86/entry/syscalls/syscall_64.tbl line 375</summary>
<p>

```
<<<<<<< ours
451	common	cachestat		sys_cachestat
452	common	fchmodat2		sys_fchmodat2
453	64	map_shadow_stack	sys_map_shadow_stack
=======
451	common	mcopy			sys_mcopy
>>>>>>> theirs
```

Resolved by concatenating and setting the mcopy line to 454.

</p>
</details>

<details><summary>include/linux/syscalls.h line 786</summary>
<p>

```c
<<<<<<< ours
/* CONFIG_MMU only */
||||||| base
/* mm/, CONFIG_MMU only */
=======
/* mm/maccess.c */
asmlinkage long sys_mcopy(void * __user dest, void * __user src, size_t len);

/* mm/, CONFIG_MMU only */
>>>>>>> theirs
```

Resolved by taking the side that added the function, but removed the (portions of) the comments that note source files.

</p>
</details>

<details><summary>include/uapi/unistd.h line 820</summary>
<p>

```c
<<<<<<< ours
#define __NR_cachestat 451
__SYSCALL(__NR_cachestat, sys_cachestat)

#define __NR_fchmodat2 452
__SYSCALL(__NR_fchmodat2, sys_fchmodat2)

#undef __NR_syscalls
#define __NR_syscalls 453
=======
#define __NR_mcopy 451
__SYSCALL(__NR_mcopy, sys_mcopy)

#undef __NR_syscalls
#define __NR_syscalls 452
>>>>>>> theirs
```

Resolved by concatenating the defines, setting mcopy's define from 451 
to 454, and setting syscalls to 455.

</p>
</details> 

### ~e65d593a18e7 -> 6b57204e1475~ dropped e65d593a18e7

<details><summary>drivers/usb/dwc3/core.c line 317</summary>
<p>

```c
<<<<<<< ours
	/*
[...]

PR #139, opened 2023-10-30
Merge tag 'v6.1.59-rt16' into nilrt/master/6.1

xhuff:
This is a regular stable merge of latest linux-stable tag v6.1.59-rt16.

[AB#2484328](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2484328)

1 merge conflict resolved:
Resolved conflict between commit e65d593 and upstream commit 3571910

Testing
* [x] boot tested on PXI VM
* [x] re-versioned NI drivers (after installing ni-daqmx) on a PXI VM via dkms autoinstall and verified there are no errors.

Additional functional and performance testing will be handle by ATSs.

@ni/rtos


drivers/usb/dwc3/core.c
```diff
@@ -311,9 +311,46 @@ int dwc3_core_soft_reset(struct dwc3 *dwc)
...
+
+		usb2_port &= ~DWC3_GUSB2PHYCFG_PHYSOFTRST;
+		dwc3_writel(dwc->regs, DWC3_GUSB2PHYCFG(0), usb2_port);
+
+		/* Wait for clock synchronization */
+		msleep(50);
+		return 0;
+	}
+
 	if (dwc->has_dsm_for_softreset) {
```
bstreiff, line 354: You resolved this merge differently than @mike-petersen-ni did in #138; he put the `if (dwc->has_dsm_for_softreset)` conditional before the new `if (dwc->dr_mode == USB_DR_MODE_HOST)` check.

I think the way you have it might be the correct one: we had `if (dwc->has_dsm_for_softreset)` after:
```c
if (dwc->dr_mode == USB_DR_MODE_HOST || dwc->current_dr_role == DWC3_GCTL_PRTCAP_HOST)
    return 0;
```
and then upstream split that into
```c
if (dwc->current_dr_role == DWC3_GCTL_PRTCAP_HOST)
    return 0;
if (dwc->dr_mode == USB_DR_MODE_HOST) {
   /*do a bunch of stuff*/ 
   return 0;
}
```
and so I think `if (dwc->has_dsm_for_softreset)` should still go after that whole block?

I'm in agreement with @gratian though on the subject of the patch we're carrying; it's an out-of-tree thing we got out of a Yocto tree and doesn't seem to have made it upstream at all, and I also wonder if it's still necessary.

bstreiff (approved):
skimming the rest of the changes I'm not seeing anything that jumps out as a concern.

drivers/usb/dwc3/core.c
xhuff: That was my rationale for positioning, yea. A previous conditional with an OR was split in two. Do you agree (given your approval) that considering dropping our patch is out of scope of this merge? Ideally it'd be dropped explicitly, or as part of the rebase?

bstreiff: Yeah, dropping it is out-of-scope for the merge; we'd want to drop it explicitly (after doing proper testing to see if we still need it or not)

PR #164, opened 2024-07-08
Merge tag 'v6.1.96-rt35' into nilrt/master/6.1

rajendra-desai-ni:
This is a regular x64 stable-rt kernel upgrade on v6.1.
Latest stable release available was v6.1.96-rt35.

[AB#2755003](https://dev.azure.com/ni/DevCentral/_workitems/edit/2755003/)

There was a merge conflict between [13dc7b7](https://github.com/ni/linux/commit/13dc7b7f158917dec94d79bdd4abea66f4f7e6ae) ("mmc: sdhci: Handle tuning error interrupts") and [b385566](https://github.com/torvalds/linux/commit/b3855668d98cf9c6aec2db999dd27d872f8ba878) ("mmc: sdhci: Add support for "Tuning Error" interrupts").

### drivers/mmc/host/sdhci.h
image

Reverted [13dc7b7](https://github.com/ni/linux/commit/13dc7b7f158917dec94d79bdd4abea66f4f7e6ae) ("mmc: sdhci: Handle tuning error interrupts") and then did git merge to include upstream changes and there were no merge conflicts.

### Testing
Rebuilt the kernel with new changes and installed on a VM and found no issues.


chaitu236:
> There was a merge conflict:

Please mention (in both PR description and commit message) which upstream commit and which existing commit from our repo conflicted. You can refer to [this PR](https://github.com/ni/linux/pull/161) as an example

image

image

chaitu236 (changes requested):
> There was a merge conflict:

Please mention (in both PR description and commit message) which upstream commit and which existing commit from our repo conflicted. You can refer to https://github.com/ni/linux/pull/161 as an example

image

image

drivers/mmc/host/sdhci.c
```diff
@@ -3627,7 +3633,7 @@ static irqreturn_t sdhci_irq(int irq, void *dev_id)
 		if (intmask & SDHCI_INT_RETUNE)
 			mmc_retune_needed(host->mmc);
 
-		if (intmask & SDHCI_INT_TUNING_ERR) {
+		if (intmask & SDHCI_INT_TUNING_ERROR) {
```
gratian: I believe our out-of-tree patch https://github.com/ni/linux/commit/13dc7b7f158917dec94d79bdd4abea66f4f7e6ae and the new upstream patch https://github.com/torvalds/linux/commit/b3855668d98cf9c6aec2db999dd27d872f8ba878 essentially accomplish the same thing, i.e. re-tune if a SDHCI_INT_TUNING_ERROR interrupt is received:

```
if (intmask & SDHCI_INT_TUNING_ERROR) {
	u16 ctrl2 = sdhci_readw(host, SDHCI_HOST_CONTROL2);

	ctrl2 &= ~SDHCI_CTRL_TUNED_CLK;
	sdhci_writew(host, ctrl2, SDHCI_HOST_CONTROL2);
}

```

By keeping both patches you will end-up triggering the re-tune twice, once from our out-of-tree patch here and before this via the newly added sdhci_data_irq() code called from a couple of lines above.

We should double check, but it looks likely we can drop our out-of-tree patch in favor of the patch that's already upstream. Our patch has an extra check and error print but I don't think that's consequential.

My recommendation would be to see if we can revert our out-of-tree patch: https://github.com/ni/linux/commit/13dc7b7f158917dec94d79bdd4abea66f4f7e6ae, do the upstream merge (which should be conflict free after revert) and test if the original problem is solved by the upstream patch. The original author of our out-of-tree patch, Kyle Roeschley should be able to help. It would be great if we don't have to carry this patch anymore.

gratian:
A minor nitpick I've noticed - when you use tags in commit messages like this:

    [rajendra-desai-ni: fix conflict https://github.com/ni/linux/commit/13dc7b7f158917dec94d79bdd4abea66f4f7e6ae with upstream commit https://github.com/torvalds/linux/commit/b3855668d98cf9c6aec2db999dd27d872f8ba878
    by updating existing MACRO_NAME SDHCI_INT_TUNING_ERR
    to upstream MACRO_NAME SDHCI_INT_TUNING_ERROR]

the usual convention is to refer to commits using the first 12 characters of the SHA-1 commit ID and the commit subject line enclosed in ("...") instead of http links.

So for example referring to this commit:

`[rajendra-desai-ni: fix conflict https://github.com/ni/linux/commit/13dc7b7f158917dec94d79bdd4abea66f4f7e6ae ...`

should be:

`[rajendra: fix conflict with 13dc7b7f1589 ("mmc: sdhci: Handle tuning error interrupts") ...`

Also it's fine to use just your first name instead of the full github username to make the tag shorter. Your full name is contained in the 'Signed-off-by' line right below that tag anyway. The goal is to be as concise as possible and ideally keep it to one line that very briefly describes what you've fixed.

drivers/mmc/host/sdhci.c
chaitu236: Btw, for WI: [AB#2316627](https://dev.azure.com/ni/DevCentral/_workitems/edit/2316627), we setup the cRIO-9053 in our rack with a swissbit card that's required to reproduce some of the bugs in [this mmc patch series](https://review-board.natinst.com/r/236135/).

The card seems to no longer plugged into the 9053 but it's probably somewhere around it - one of our technicians can help plug it in.

rajendra-desai-ni: I am able to revert the out-of-tree patch and do the upstream merge and did not encounter any conflicts this time. I am able to rebuild the kernel and did a basic sanity check, but not sure how to test only those specific changes that we pulled from [upstream commit](https://github.com/torvalds/linux/commit/b3855668d98cf9c6aec2db999dd27d872f8ba878). I have pinged Kyle Roeschley to get help for testing the changes. Will update here once the new changes work.

gratian:
Closing this PR for now until we work out a fix for the SDHCI conflict which would be acceptable upstream.
Also `v6.1.99-rt36` has since become available upstream so we'll want an updated PR anyway.

PR #167, opened 2024-08-20
Merge tag 'v6.1.105-rt38' into nilrt/master/6.1

erickshepherdNI:
There were no merge conflicts.

### TESTING
- [X] Built the kernel locally
- [X] Installed the new kernel on a cRIO-9033
- [X] Confirmed the newly installed kernel version is 6.1.105-rt38
- [X] Re-versioned drivers via DKMS and confirmed the 6.1.105-rt38 versions are installed


chaitu236 (commented):
Why are there two merge commits and one of them looking empty (and unsigned)?

PR #165, opened 2024-08-15
Merge tag 'v6.1.99-rt36' into nilrt/master/6.1

erickshepherdNI:
Fix one conflict with https://github.com/erickshepherdNI/linux/commit/32b76505ba18cbbfd7c41f7962937bea5481502c ("mmc: sdhci: Add support for "Tuning Error" interrupts") in favor of our https://github.com/erickshepherdNI/linux/commit/13dc7b7f158917dec94d79bdd4abea66f4f7e6ae ("mmc: sdhci: Handle tuning error interrupts"). This was done in a separate commit after the merge.

### TESTING

- [X] Built the kernel locally
- [X] Installed the new kernel on a cRIO-9033
- [X] Confirmed the newly installed kernel version is 6.1.99-rt36

PR #170, opened 2024-08-23
nati_x86_64_defconfig: enable nftables support

rtollert:
Manually enabled CONFIG_NFT_* and selected CONFIG_NF_* and CONFIG_IP*_NF_* options via menuconfig.

Somewhere between menuconfig and savedefconfig, CONFIG_DRM_VMWGFX_FBCON=y was removed. I didn't do that.

Testing: rebuilt nilrt-base-system-image; haven't attempted runtime testing.


gratian (commented):
The change is fine, the commit message could use some improvement: it is missing the "why?" It doesn't have to be too verbose.

Something like "Enable COFNIG_NFT ... for <blah> product/configuration/etc." should be sufficient for somebody to know later why we enabled them and if it is safe to change them.

For the 'CONFIG_DRM_VMWGFX_FBCON=y' diff I would recommend creating a separate commit where you just regenerate the defconfig (before making your nftables changes). The commit description can be just "Regenerate. No functional changes."  (with the usual subject/signoff).

gratian:
Also can you add a "Testing" section to your PR where you describe what testing you've done if any. 

I would also like to see the disk size increase this adds. It don't expect it to gate the PR but it would be good to know for future reference.

rtollert:
> I would also like to see the disk size increase this adds. It don't expect it to gate the PR but it would be good to know for future reference.

nilrt-base-system-image-x64.tar:
- latest autobuild: 282.685 KB.
- test build with my branch: 282.736 KB (+0.02%).

Note that I haven't actually added any kernel modules, userspace packages etc. to the packagegroups yet.

rtollert:
> For the 'CONFIG_DRM_VMWGFX_FBCON=y' diff I would recommend creating a separate commit where you just regenerate the defconfig (before making your nftables changes). The commit description can be just "Regenerate. No functional changes." (with the usual subject/signoff).

But shouldn't the defconfigs always be regenerated rather than hand-edited? In any case, I'll split up the commit while trying to connote that's what's going on.

rtollert:
@gratian v2 is up and I think I've addressed everything?

gratian:
> > For the 'CONFIG_DRM_VMWGFX_FBCON=y' diff I would recommend creating a separate commit where you just regenerate the defconfig (before making your nftables changes). The commit description can be just "Regenerate. No functional changes." (with the usual subject/signoff).
> 
> But shouldn't the defconfigs always be regenerated rather than hand-edited? In any case, I'll split up the commit while trying to connote that's what's going on.

Yes, defconfigs should always be regenerated and not hand-edited. In this case what it appears happened is commit 67adcfae2eb177303ceba1fbcb0935909c6fecfa ("drm/vmwgfx: Port the framebuffer code to drm fb helpers") removed the `DRM_VMWGFX_FBCON` config option. We picked it up with the upstream merge `v6.1.99-rt36` in PR https://github.com/ni/linux/pull/165 but neglected to regenerate our defconfig. I will make sure our future work templates will call out checking the nati defconfig after merges. 

That's why I was saying above you should regenerate the defconfig **before** you make the other changes. With the split commits the order is reversed. I'll fix it when I pull it.

gratian:
> > I would also like to see the disk size increase this adds. It don't expect it to gate the PR but it would be good to know for future reference.
> 
> nilrt-base-system-image-x64.tar:
> 
>     * latest autobuild: 282.685 KB.
> 
>     * test build with my branch: 282.736 KB (+0.02%).
> 
> 
> Note that I haven't actually added any kernel modules, userspace packages etc. to the packagegroups yet.

Checking the actual modules folder in a local build:
```
/lib/modules$ du -cs *
46948   6.1.105-rt38-00106-ga86ee30b566a
47996   6.1.105-rt38-00108-g616ddd777a62
```
A little over 1 MB.

Note that currently all enabled kernel modules get installed by packagegroup-ni-base: https://github.com/ni/meta-nilrt/blob/805579d6d77a08630f25273274cc50a03c017523/recipes-core/packagegroups/packagegroup-ni-base.bb#L57

We do have plans to add optional modules but they are still in the planning stage.

gratian:
Merge completed outside GitHub UI. Reversed order of commits as described above.

```
f947d0d0bdb86ea86173ce966c3adf86317e0ec8 (HEAD -> nilrt/master/6.1, origin/nilrt/master/6.1) nati_x86_64_defconfig: enable nftables support
8309134753e01d2074fc977785e6f3ee41b4ced1 nati_x86_64_defconfig: regenerate; no functional changes
a86ee30b566af1563ceca9e69d5d99144f57cd09 (rtollert/nilrt/master/6.1, erickshepherd/nilrt/master/6.1) Merge tag 'v6.1.105-rt38' into nilrt/master/6.1

```

Verified there are no superfluous changes after each commit by regenerating the defconfig during the rebase.

gratian:
Cherry-picked in `nilrt/master/6.6`.

PR #171, opened 2024-08-23
zynq: nftables support

rtollert:
This more or less mirrors #170, but I also added a build fix for newer gccs that came up when trying to use `make menuconfig`. I imagine that this was already fixed in upstream by 6.1 but I didn't want to deal with the rigamarole of a backport.


rtollert:
@amstewart Looks like `nilrt/master/sumo` is building from `nilrt/22.8/4.14`, *not* `nilrt/master/4.14`. Does it need to be updated? Or do I need to re-submit this PR to a separate branch? Or do I need to cut a new `nilrt/24.8/sumo` branch?

rtollert:
Withdrawn (after talking to @amstewart offline).

PR #175, opened 2024-09-16
nati_x86_64_defconfig: enable CONFIG_NFT_FIB_* and CONFIG_NLMON

rtollert:
It was observed that firewalld was failing on startup with the error

ERROR: COMMAND_FAILED: 'python-nftables' failed: internal:0:0-0: Error: Could
not process rule: No such file or directory

I used nlmon to debug this by capturing and partially decoding the netlink traffic in wireshark. On that basis, I think setting CONFIG_NLMON=m was prudent.

That analysis implicating something fib-related in the nft kernel config. Those modules weren't enabled; enabling them eliminates the error.

I'm also setting CONFIG_NFT_DUP_*=m to proactively avoid any potential mysterious firewalld failures due to missing modules in the future.


gratian (approved):

gratian:
For future reference the disk size impact of adding these modules is 72 KBytes:

```
$ du -cs *
48052   6.1.108-rt40-00110-gcd807a6581c8
48124   6.1.108-rt40-00111-g853ad5b299f8
```

gratian:
@rtollert I will cherry pick this in our 'next' kernel branch (6.6).

Please note that the release branches for this cycle have already been created (i.e. nilrt/24.8/6.1). Let me know if this change needs to be cherry picked into the 24.8 release. Based on the fact that the rest of firewald changes are being staged in distro 'next' I'm going to assume the answer is no.

PR #181, opened 2024-11-13
Update error handling for SDHCI tuning

erickshepherdNI:
This pull request reverts our previous commit to handle tuning interrupts in favor of the upstream commit that does the same thing. 

This pull request also includes a fix for the sdhci_execute_tuning function, which was not actually returning the result of the underlying tuning function. This caused multiple retuning attempts for DDR50 cards, which cannot always be tuned, that led to an I/O error. This change updates sdhci_execute_tuning to return the error code which causes tuning to be disabled after the first attempt.

### Testing
I built the kernel and installed it on two different setups that both use DDR50 SD cards. Previously only one setup reproduced the I/O error that resulted from repeated retuning. After installing the kernel both systems reported that the initial DDR50 tuning failed, but no other issues appeared in the dmesg logs and I was able to read and write to the SD cards on both systems.


erickshepherdNI:
I'm continuing to investigate this issue further and may need to make additional changes. I've converted this to a draft until it is ready to go in.

gratian:

PR #190, opened 2025-01-21
Ershephe/update tuning error handling 6.6 nr

erickshepherdNI:
Fixes cases in the MMC driver where an SD card incapable of tuning is requested to tune more than once, which would result in I/O errors. This required a change in sdhci_execute_tuning() to return the error code returned by  __sdhci_execute_tuning(), which disables any further tuning. I believe this error not being returned is an oversight since the error from platform_execute_tuning() in the same function is returned this way and the calling function, mmc_sd_init_uhs_card(), has an if statement handling the returned error for the DDR50 card we are using.

The new field in the mmc_host struct is used to prevent a card incapable of tuning from being tuned during reinitialization. This prevents I/O errors for DDR50 cards that are reset for any reason. Currently this field is only set for DDR50 cards that fail to tune to limit the impact of these changes.

Commits to revert our fix for tuning error interrupts and reapply the upstream fix are also included.

[AB#2805437](https://dev.azure.com/ni/DevCentral/_workitems/edit/2805437)

# Testing:
I tested these changes on a cRIO 9053 with a DDR50 SD card by building the linux kernel locally and applying it to the device. I confirmed that no I/O errors were thrown, even upon a hardware reset of the card, and that the card was properly mounted and could be written to.

Fixes I/O errors when using a DDR50 SD card that does not support tuning.


drivers/mmc/core/sd.c
```diff
@@ -668,7 +668,8 @@ static int mmc_sd_init_uhs_card(struct mmc_card *card)
 	if (!mmc_host_is_spi(card->host) &&
 		(card->host->ios.timing == MMC_TIMING_UHS_SDR50 ||
 		 card->host->ios.timing == MMC_TIMING_UHS_DDR50 ||
-		 card->host->ios.timing == MMC_TIMING_UHS_SDR104)) {
+		 card->host->ios.timing == MMC_TIMING_UHS_SDR104) &&
+		 !card->host->never_retune) {
```
chaitu236: minor: I'm not entirely certain of formatting rules for kernel, but since `!card->host->never_retune` isn't part of the previous lines' parenthesis, shouldn't `!` align with `(` on line 669?

drivers/mmc/core/sd.c
```diff
@@ -681,6 +682,7 @@ static int mmc_sd_init_uhs_card(struct mmc_card *card)
 		if (err && card->host->ios.timing == MMC_TIMING_UHS_DDR50) {
 			pr_warn("%s: ddr50 tuning failed\n",
 				mmc_hostname(card->host));
+			card->host->never_retune = 1;
```
chaitu236: Did you check that the only reason `mmc_execute_tuning` can fail is when the card doesn't support tuning?
Or can there be a situation where a card is capable of tuning but fails tuning for another reason or in just one instance?

chaitu236 (commented):
Since the patch would be submitted to upstream master, more specifically the development tree of whoever maintains ~these patches~ this part of codebase, did you check if these changes apply cleanly on that tree and build?

drivers/mmc/core/sd.c
erickshepherdNI: I think you're right. I'll update the spacing.

drivers/mmc/core/sd.c
erickshepherdNI: mmc_execute_tuning has two error codes it can return. One is a timeout code and the other is a try again code. I'm not sure if the only way the tuning can time out is if it is unsupported but that seems to be the assumption in place for this if-statement since it was added to handle the case when tuning is not supported on some DDR50 cards by clearing the error.

I could update this to only set the never_retune field in the case of a timeout error code. It may even be correct to apply that to the whole if-statement since I'm not sure if we should be clearing a try again code. What do you think?

drivers/mmc/host/sdhci.h
```diff
@@ -158,6 +158,7 @@
 #define  SDHCI_INT_BUS_POWER	0x00800000
 #define  SDHCI_INT_AUTO_CMD_ERR	0x01000000
 #define  SDHCI_INT_ADMA_ERROR	0x02000000
+#define  SDHCI_INT_TUNING_ERROR	0x04000000
```
gratian, line 161: The define for SDHCI_INT_TUNING_ERROR should be in the commit that reapplies the upstream patch, right?

It is not something you've added so it shouldn't show up in this diff. AFAIK it was introduced by the upstream commit: https://github.com/ni/linux/commit/b3855668d98cf9c6aec2db999dd27d872f8ba878

erickshepherdNI: Good catch. I remember this change was left out when I reverted our fix and reapplied the upstream one so I must have fixed it after and included it in this commit. I can try to rework it to be in the other commit and will not include it in the upstream patch.

drivers/mmc/core/sd.c
chaitu236: > since it was added to handle the case when tuning is not supported on some DDR50 cards by clearing the error.

Do we know this for certain? Couldn't it also be because the tuning may sometimes fail and retrying would fix it?
I guess even if we're not sure about this, we could send this upstream and see if we get any pushback.

erickshepherdNI: That could also be the case. Unfortunately the [commit](https://github.com/ni/linux/commit/4324f6de6d2eb9b232410eb0d67bfafdde8ba711) that added the if-statement, as well as tuning for DDR50 cards, doesn't specify it's exact purpose and the comment above it doesn't either. This is the change I'm most unsure about but I think sending it upstream and including the people who added the commits that enabled DDR50 tuning will provide more clarity.

erickshepherdNI:
> Since the patch would be submitted to upstream master, more specifically the development tree of whoever maintains ~these patches~ this part of codebase, did you check if these changes apply cleanly on that tree and build?

The changes do apply cleanly and I was able to build the upstream kernel with them. I've attached the patch files I tested to the original user story and can get them sent upstream today.

chaitu236:
> > Since the patch would be submitted to upstream master, more specifically the development tree of whoever maintains ~these patches~ this part of codebase, did you check if these changes apply cleanly on that tree and build?
> 
> The changes do apply cleanly and I was able to build the upstream kernel with them. I've attached the patch files I tested to the original user story and can get them sent upstream today.

GitHub's UI doesn't make this obvious but just noticed this from the patch files - commit msg line lengths seem a little too long.

Try to follow [50/72 rule](https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting) i.e., first line 50 chars, followed by new line followed by lines within 70-75 chars. You may also want to check https://docs.kernel.org/process/submitting-patches.html.

gratian:

PR #191, opened 2025-01-26
mtd: Cherry pick MTD related commits for Bluefin

HatsyRei:
WI: [AB#2959883](https://dev.azure.com/ni/DevCentral/_workitems/edit/2959883/)

Part of work to get 6.6 kernel support for Bluefin. Cherry pick MTD-related commits applicable to Bluefin. 

Testing: Manually built kernel with cherry picked commits.

### DIFF
--- a/drivers/mtd/nand/raw/pl35x-nand-controller.c
+++ b/drivers/mtd/nand/raw/pl35x-nand-controller.c
@@ -28,7 +28,7 @@
 #include <linux/slab.h>
 #include <linux/clk.h>
 
-#define PL35X_NANDC_DRIVER_NAME "pl35x-nand-controller"
+#define PL35X_NANDC_DRIVER_NAME "xilinx_nand"
 
 /* SMC controller status register (RO) */
 #define PL35X_SMC_MEMC_STATUS 0x0
--- a/drivers/mtd/parsers/Kconfig
+++ b/drivers/mtd/parsers/Kconfig
@@ -67,6 +67,30 @@ config MTD_CMDLINE_PARTS
 
 	  If unsure, say 'N'.
 
+config MTD_RESERVE_END
+	int "Reserved space at the end of an all remaining space partition"
+	depends on MTD_CMDLINE_PARTS = "y"
+	default 0
+	help
+	  Specify an amount of reserved space at the end of the last MTD
+	  partition when the size is specified with '-' to denote all
+	  remaining space.
+
+	  This can be useful if, for example, the BBT is stored at the end
+	  of the flash, and you don't want those blocks counted as part of
+	  the last MTD partition. This is less heavyweight than reserving
+	  the BBT blocks with a separate MTD partition. The BBT marks its
+	  own blocks as bad blocks, which prevents an MTD driver such as
+	  UBI from getting an accurate count of the actual bad blocks in
+	  the MTD partition that contains the BBT.
+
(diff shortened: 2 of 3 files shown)


drivers/mtd/ubi/debug.c
```diff
@@ -556,7 +556,7 @@ int ubi_debugfs_init_dev(struct ubi_device *ubi)
 						   mode, d->dfs_dir,
 						   (void *)ubi_num, &dfs_fops);
 
-	debugfs_create_file("detailed_erase_block_info", S_IRUSR, d->dfs_dir,
+	debugfs_create_file("detailed_erase_block_info", S_IRUGO, d->dfs_dir,
```
chaitu236: Why is this needed on bluefin?

HatsyRei: @chaitu236 There's an archived escalation ([link](https://emerson.sharepoint.com/sites/TM-industrial-embedded-pse/SitePages/Track%20sbRIO-9651%20SSD%20Health.aspx)) about tracking flash memory health. It appears that the original intent of the commit is to allow customers to access `detailed_erase_block_info` through all interfaces available to them (which they otherwise would not be able to due to file permission issues).

Given that code search did not locate any internal components which pull data from this file, I'm inclined to believe that this commit is not specific to Bluefin, nor do any Bluefin components depend on it. Would you advise I remove this commit?

chaitu236: Not sure if there'd be security concerns with making this file more permissive, but if upstream has chosen S_IRUSR, I'd prefer to keep it that way, especially since we haven't had this commit on x64 targets for many years now and no one complained.

But I'd leave it to @gratian to decide it.

chaitu236 (approved):

drivers/mtd/ubi/debug.c
gratian: > Given that code search did not locate any internal components which pull data from this file, I'm inclined to believe that this commit is not specific to Bluefin, nor do any Bluefin components depend on it. Would you advise I remove this commit?

Yes. Please drop this commit.

Nothing uses it - there's no reason to carry an out-of-tree commit with potential security implications. An end-user can use appropriate privileged user accounts to access the erase block information and/or modify the access permissions themselves (in user-space) if so desired.

HatsyRei: Got it. Dropped.

drivers/mtd/nand/raw/pl35x-nand-controller.c
```diff
@@ -28,7 +28,7 @@
 #include <linux/slab.h>
 #include <linux/clk.h>
 
-#define PL35X_NANDC_DRIVER_NAME "pl35x-nand-controller"
+#define PL35X_NANDC_DRIVER_NAME "xilinx_nand"
```
HatsyRei, line 31: Cherry picked one more commit (https://github.com/ni/linux/commit/df465252df84668ac5318fe6394eca5e3fe38426) as we're observing the same issue with driver name not lining up with what u-boot provides as boot arguments for partitioning the nand.

gratian: Gah, this is coming back to haunt me. I really wish we would have fixed it in u-boot at the time. I guess we don't have a choice now and have to keep this around.

gratian (approved):

PR #192, opened 2025-01-30
ARM: dts: Update Bluefin device trees

HatsyRei:
WI: [AB#2959883](https://dev.azure.com/ni/DevCentral/_workitems/edit/2959883/)

Changes required to build ni-bluefin.dts with updated [zynq-7000.dtsi](https://github.com/ni/linux/blob/nilrt/master/6.6/arch/arm/boot/dts/xilinx/zynq-7000.dtsi) device tree bindings.

Testing: Manually built kernel with changes and decompiled resulting ni-bluefin.dtb via `dtc` to ensure changes are applied.

### DIFF
--- a/arch/arm/boot/dts/xilinx/ni-zynq.dtsi
+++ b/arch/arm/boot/dts/xilinx/ni-zynq.dtsi
@@ -138,6 +138,20 @@
 	};
 };
 
+&smcc {
+	status = "okay";
+};
+
+&nfc0 {
+	status = "okay";
+
+	nand@0 {
+		reg = <0>;
+		nand-on-flash-bbt;
+		nand-ecc-mode = "on-die";
+	};
+};
+
 &L2 {
 	prefetch-data = <1>;
 	prefetch-instr = <1>;
--- a/arch/arm/boot/dts/xilinx/ni-zynq.dtsi
+++ b/arch/arm/boot/dts/xilinx/ni-zynq.dtsi
@@ -105,19 +105,6 @@
 			clock-frequency = <58824000>;
 		};
 
-		i2c0: i2c@e0004000 {
-			nicpld@40 {
-				compatible = "ni,cpld";
-				reg = <0x40>;
-			};
-
-			ds3231_rtc@68 {
-				compatible = "ds3232";
-				status = "disabled";
-				reg = <0x68>;
-			};
-		};
-
 		i2c2: i2c@81000000 {
 			compatible = "xlnx,xps-iic-2.00.a";
 			status = "disabled";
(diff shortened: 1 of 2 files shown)


arch/arm/boot/dts/xilinx/ni-zynq.dtsi
```diff
@@ -124,108 +105,6 @@
...
-		};
-
-		can1: can@e0009000 {
-			compatible = "xlnx,zynq-can-1.00.a";
-			status = "disabled";
-			reg = <0xe0009000 0x1000>;
-			interrupts = <0 51 4>;
-		};
-
-		fpgaperipheral@f8007000 {
```
gratian, line 141: I don't see 'fpgaperipheral' listed in the parent zynq-7000.dtsi. I guess the base address matches 'devcfg@f8007000' so removing it here may be the right thing but I can't tell since it's not listed in the commit message.

arch/arm/boot/dts/xilinx/ni-zynq.dtsi
```diff
@@ -235,38 +114,41 @@
...
-			clocks = <&clkc 26>, <&clkc 35>;
-			#address-cells = <1>;
-			#size-cells = <0>;
-		};
+&smcc {
+	status = "okay";
+};
 
-		usb0: usb@e0002000 {
-			usb-phy = <&usb_phy0>;
```
gratian, line 262: I don't think the usb-phy attribute comes from zynq-7000.dtsi since the usb_phy0 node is declared in this file, see lines 14-17. I suspect you need to keep this.

arch/arm/boot/dts/xilinx/ni-zynq.dtsi
```diff
@@ -235,38 +114,41 @@
...
-		};
+&nfc0 {
+	status = "okay";
 
-		usb1: usb@e0003000 {
-			usb-phy = <&usb_phy1>;
```
gratian, line 266: Same comments as above. In general, make sure all the stuff you are removing actually has an equivalent in the common zynq-7000.dtsi we are including and is not just being extended here.

arch/arm/boot/dts/xilinx/ni-zynq.dtsi
HatsyRei: The devcfg node does appear to represent the Xilinx FPGA manager ([ref](https://www.kernel.org/doc/Documentation/devicetree/bindings/fpga/xilinx-zynq-fpga-mgr.txt)). Will amend the commit message to clarify this.

arch/arm/boot/dts/xilinx/ni-zynq.dtsi
```diff
@@ -235,38 +114,41 @@
...
+	};
+};
+
+&usb0 {
+	usb-phy = <&usb_phy0>;
+};
 
+&usb1 {
+	usb-phy = <&usb_phy1>;
 };
```
HatsyRei: @gratian Regarding your comments on the usb-phy attributes, I modified them to be added to the usb nodes by referencing the usb phandles. Should we keep the original bindings?

gratian: OK. I think that's fine but it wasn't immediately obvious why the got removed.

That tells me you should probably split this into its own commit with a commit message that explains why they are not needed anymore (i.e. they're "added to the usb nodes by referencing the usb phandles ...")

gratian:
Also, can you rebase this PR on the latest nilrt/master/6.6. Sorry, I've pulled another PR before yours (kernel upgrade).

arch/arm/boot/dts/xilinx/ni-zynq.dtsi
HatsyRei: Split these (along with the same change for i2c0) into a separate commit.

gratian (approved):

PR #194, opened 2025-02-05
Resolve I/O errors for DDR50 cards

erickshepherdNI:
Fixes cases in the MMC driver where an SD card incapable of tuning is requested to tune more than once, which would result in I/O errors. This required a change in sdhci_execute_tuning() to return the error code returned by  __sdhci_execute_tuning(), which disables any further tuning. I believe this error not being returned is an oversight since the error from platform_execute_tuning() in the same function is returned this way and the calling function, mmc_sd_init_uhs_card(), has an if statement handling the returned error for the DDR50 card we are using.

The new field in the mmc_card struct is used to prevent a card incapable of tuning from being tuned again. This prevents I/O errors for DDR50 cards that are reset for any reason. Currently this field is only set for DDR50 cards that fail to tune to limit the impact of these changes.

Commits to revert our fix for tuning error interrupts and reapply the upstream fix are also included.

This PR is nearly identical to #190 but moves the field used to disable tuning to the mmc_card struct based on feedback from upstream. I tried to get a working solution without the change to return the error code, but each solution I tried caused the cRIO to stall. I suspect trying to write to the card struct within the sdhci functions was the issue but I'm still not sure.

[AB#2805437](https://dev.azure.com/ni/DevCentral/_workitems/edit/2805437)


gratian (approved):
Looks good to me. We'll find out from upstream if that's the correct structure to add the new field in (they may want one of the other structs included there but it's hard to say w/o their input) and what they want to do about returning the error.

I suggest running '<linux_source>/scripts/get_maintainer.pl' on the patches to figure out who needs to be included on the email.

chaitu236 (approved):

gratian:

PR #196, opened 2025-02-13
Update serial_core.c to create an anonymous parent device if parent device pointer is null

nikhalim:
niserial987xDriver doesn't pass in a parent device pointer to uartPort when calling uart_add_one_port. This, in turn, crashes the kernel when serial_core attempts to locate the parent device pointer (since the pointer is NULL).

This PR attempts to use platform_device_register_simple to create a platform device that could be assigned as the parent device for NI-987x ports. It does this by checking if the caller is niserial987xDriver. If it is, it gets the slot number of the 987x module, and creates an appropriate platform_device for each uart port that needs to be registered.

### DIFF
--- a/drivers/tty/serial/serial_core.c
+++ b/drivers/tty/serial/serial_core.c
@@ -31,6 +31,7 @@
 
 #include <linux/irq.h>
 #include <linux/uaccess.h>
+#include <linux/platform_device.h>
 
 #include "serial_base.h"
 
@@ -3404,6 +3405,22 @@ int serial_core_register_port(struct uart_driver *drv, struct uart_port *port)
 	 */
 	port->flags |= UPF_DEAD;
 
+	if (port->dev == NULL) {
+		static unsigned int next_port_number;
+		char port_name[21];
+		struct platform_device *pdev;
+
+		snprintf(port_name, sizeof(port_name), "anon_port_%u", next_port_number++);
+		pdev = platform_device_register_simple(port_name, -1, NULL, 0);
+
+		if (PTR_ERR_OR_ZERO(pdev)) {
+			ret = -EINVAL;
+			goto err_unlock;
+		}
+
+		port->dev = &pdev->dev;
+	}
+
 	/* Inititalize a serial core controller device if needed */
 	ctrl_dev = serial_core_ctrl_find(drv, port->dev, port->ctrl_id);
 	if (!ctrl_dev) {
@@ -3440,6 +3457,9 @@ err_unregister_port_dev:
 err_unregister_ctrl_dev:
 	serial_base_ctrl_device_remove(new_ctrl_dev);
 
+	if (strncmp(to_platform_device(port->dev)->name, "anon_port", 9) == 0)
+		platform_device_del(to_platform_device(port->dev));
+
 err_unlock:
 	mutex_unlock(&port_mutex);
 
@@ -3469,6 +3489,9 @@ void serial_core_unregister_port(struct uart_driver *drv, struct uart_port *port
 	/* Drop the serial core controller device if no ports are using it */
 	if (!serial_core_ctrl_find(drv, phys_dev, ctrl_id))
 		serial_base_ctrl_device_remove(ctrl_dev);
+
+	if (strncmp(to_platform_device(phys_dev)->name, "anon_port", 9) == 0)
+		platform_device_del(to_platform_device(phys_dev));
 
 	mutex_unlock(&port_mutex);
 }


nikhalim:
@gratian @sjasonsmith I couldn't add reviewers so I'm using this comment to notify reviewers.

sjasonsmith (approved):
This seems like a reasonable workaround while further work is done to seek a better solution not requiring changes to serial_core.

I like that the changes are confined into two contiguous blocks of code, which should make it easier to deal with any merge conflicts which may arise before this can be eliminated.

You'll need approval from @gratian or others responsible for this fork to decide whether it's good to go.

gratian:
@nikhalim For this git repo (and other open source repos) we follow the upstream kernel rules when it comes to commit messages and merges. That means:
- No squash merges. Each commit/patch should be a single logical change and be self contained. It looks like you have 2 commits that should be combined into a single clean one. PR feedback should be addressed by amending the original commit and force-pushing the update to your branch. The GitHub UI knows how to handle that in an open PR.
- Please follow the upstream rules[1] when it comes to commit messages:
  - Subject line should be 50 characters or less (if possible) and include the subsytem(s) the patch applies to. Something like "serial: core: create anonymous parent devices for NI-987x" should work even though is slightly longer than the official limit.
  - Include an empty line following the subject. This is important for git short logs to look OK and be useful during merges/rebases.
  - Wrap the commit message body at 72 characters and make sure it contains the "why" this change is needed. I think you already explain the "why" in your first commit.
  - Describe your changes in imperative mood (also covered in the link below). For example don't say "This PR attempts to use ..." instead say "Use platform_device_register_simple to create a platform device that could be assigned as the parent device for NI-987x ports if the caller is niserial987xDriver".
  - Sign your commits with "Signed-off-by: ...", see the section on the "Developer’s Certificate of Origin" in the kernel documentation linked[1].

I know this may seem nit-picky but following these rules makes it much easier to maintain these commits in the future and have them work well with git tooling like bisect, rebases etc.

[1] https://www.kernel.org/doc/html/latest/process/submitting-patches.html

nikhalim:
@gratian I've already squashed and updated the commit message. Hopefully I did it right.

drivers/tty/serial/serial_core.c
```diff
@@ -3404,6 +3404,40 @@ int serial_core_register_port(struct uart_driver *drv, struct uart_port *port)
...
+				{"987x_4_1", "987x_4_2", "987x_4_3", "987x_4_4"},
+				{"987x_5_1", "987x_5_2", "987x_5_3", "987x_5_4"},
+				{"987x_6_1", "987x_6_2", "987x_6_3", "987x_6_4"},
+				{"987x_7_1", "987x_7_2", "987x_7_3", "987x_7_4"},
+				{"987x_8_1", "987x_8_2", "987x_8_3", "987x_8_4"}};
+			static struct platform_device *anonymous_parent_dev = NULL;
+			int slot_number = 0;
+			
+			if (sscanf(driver_name, "niserial987xSlot%uDriver", &slot_number) != 1){
+				goto err_unlock;
```
sjasonsmith: These are all going to cause the uninitialized `ret` value to be returned, which might randomly indicate success.

All of these `goto err_unlock` paths should set an error code, perhaps `-EINVAL`. This is the error which is eventually set by the `serial_core_ctr_device_add` if it fails further down.

sjasonsmith (changes requested):

drivers/tty/serial/serial_core.c
```diff
@@ -3470,6 +3504,11 @@ void serial_core_unregister_port(struct uart_driver *drv, struct uart_port *port
 	if (!serial_core_ctrl_find(drv, phys_dev, ctrl_id))
 		serial_base_ctrl_device_remove(ctrl_dev);
 
+	/* Remove the platform_device created for niserial987xDriver */
+	if (strncmp(drv->driver_name, "niserial987x", 12) == 0) {
+		platform_device_del(to_platform_device(phys_dev));
```
chaitu236, line 3494: Shouldn't something similar be done on line 3476 as well so `err_unregister_ctrl_dev` and `err_unregister_port_dev` codepaths would also delete the platform device.

drivers/tty/serial/serial_core.c
```diff
@@ -3404,6 +3404,40 @@ int serial_core_register_port(struct uart_driver *drv, struct uart_port *port)
 	 */
 	port->flags |= UPF_DEAD;
 
+	if (port->dev == NULL) {
+		const char *driver_name = drv->driver_name;
+		
+		if (strncmp(driver_name, "niserial987x", 12) == 0) {
+			// Max number of cRIO modules = 8
+			// Max number of uart ports per module = 4
+			static const char *const platform_device_names[8][4] = {
```
chaitu236: Is there a reason we need this array and can't use `snprintf` or such to construct device name programmatically?

chaitu236 (changes requested):

drivers/tty/serial/serial_core.c
nikhalim: The strings need to be constant as platform_device_register_simple expects const char

sjasonsmith: It actually just stores the passed-in pointer. It doesn't copy the string into its own buffer. I explored that path when I reviewed your first implementation of this.

drivers/tty/serial/serial_core.c
sjasonsmith: Yes, I think you are right. Of course in this context you could just add a local variable to track whether they were created rather then doing all the string parsing again.

drivers/tty/serial/serial_core.c
nikhalim: Assigned ret with -EINVAL

drivers/tty/serial/serial_core.c
nikhalim: Added platform_device_del to err_unregister_ctrl_dev

chaitu236 (approved):

gratian:
@nikhalim Can you rebase your branch on top of the latest `nilrt/master/6.6`. Sorry, I've pulled a PR before yours so the head changed.

More importantly can you run your patch through `<linux_source>/scripts/checkpatch.pl`? You can generate the patch from your branch with `git format-patch -1`. It is currently showing quite a few issues:

```
$ ./scripts/checkpatch.pl 0001-serial-core-create-anonymous-parent-for-NI-987x.patch
ERROR: trailing whitespace
#41: FILE: drivers/tty/serial/serial_core.c:3411:
+^I^I$

ERROR: do not initialise statics to NULL
#54: FILE: drivers/tty/serial/serial_core.c:3424:
+                       static struct platform_device *anonymous_parent_dev = NULL;

ERROR: trailing whitespace
#56: FILE: drivers/tty/serial/serial_core.c:3426:
+^I^I^I$

ERROR: space required before the open brace '{'
#57: FILE: drivers/tty/serial/serial_core.c:3427:
+                       if (sscanf(driver_name, "niserial987xSlot%uDriver", &slot_number) != 1){

ERROR: trailing whitespace
#61: FILE: drivers/tty/serial/serial_core.c:3431:
+^I^I^I$

WARNING: line length of 144 exceeds 100 columns
#62: FILE: drivers/tty/serial/serial_core.c:3432:
+                       anonymous_parent_dev = platform_device_register_simple(platform_device_names[slot_number - 1][port->line], -1, NULL, 0);

ERROR: trailing whitespace
#66: FILE: drivers/tty/serial/serial_core.c:3436:
+^I^I^I}^I^I$

ERROR: else should follow close brace '}'
#71: FILE: drivers/tty/serial/serial_core.c:3441:
+               }
+               else {

WARNING: braces {} are not necessary for single statement blocks
#85: FILE: drivers/tty/serial/serial_core.c:3484:
+       if (is_niserial987x) {
+               platform_device_del(to_platform_device(port->dev));
+       }

WARNING: braces {} are not necessary for single statement blocks
#97: FILE: drivers/tty/serial/serial_core.c:3519:
+       if (strncmp(drv->driver_name, "niserial987x", 12) == 0) {
+               platform_device_del(to_platform_device(phys_dev));
+       }

WARNING: From:/Signed-off-by: email name mismatch: 'From: Kevin Lim <khai-wern.lim@ni.com>' != 'Signed-off-by: Kevin Khai-Wern Lim <khai-wern.lim@ni.com>'

total: 7 errors, 4 warnings, 80 lines checked
```

nikhalim:
```
./scripts/checkpatch.pl 0001-serial-core-create-anonymous-parent-for-NI-987x.patch
total: 0 errors, 0 warnings, 80 lines checked

0001-serial-core-create-anonymous-parent-for-NI-987x.patch has no obvious style problems and is ready for submission.
```

Fixed the styling.

drivers/tty/serial/serial_core.c
gratian: @sjasonsmith like @chaitu236 that ugly array rubs me the wrong way so I went digging through the code and unless I'm missing something I can see space being allocated and the name being copied. The code path is:

```
platform_device_register_simple()
	platform_device_register_resndata(.., name, ...)
		struct platform_device_info pdevinfo = {
			.name = name,
			...
		}
                platform_device_register_full(&pdevinfo)
		        platform_device_alloc(pdevinfo->name, pdevinfo->id);
			        pa = kzalloc(sizeof(*pa) + strlen(name) + 1, GFP_KERNEL); 
			        strcpy(pa->name, name);
			        pa->pdev.name = pa->name;
```

It is true that the passed-in pointer gets stored in platform_device_register_resndata() but that's just temporarily in a structure on the stack.

The other evidence we have that we don't need to keep a variable storing the string around is the fact that there are are many examples in the kernel where platform_device_register_simple() is called with a string literal.

@nikhalim can you give this commit a try: https://github.com/gratian/linux/commit/4a49adcd6ca30cde1073508d0af811254670feb0

This reuses the driver_name for the platform_device_register_simple(). As far as I can tell that should be OK but if we do need to create unique names per port (and not just slot) then we can go the `kzalloc + snprintf` route.

nikhalim: Following @sjasonsmith suggestion, this is what it looks like now:
```
static int next_port_number = 0;
char port_name[21];
struct platform_device *pdev;

snprintf(port_name, sizeof(port_name), "anon_port_%d", next_port_number++);
pdev = platform_device_register_simple(port_name, -1, NULL, 0);
```

drivers/tty/serial/serial_core.c
```diff
@@ -3404,6 +3405,22 @@ int serial_core_register_port(struct uart_driver *drv, struct uart_port *port)
 	 */
 	port->flags |= UPF_DEAD;
 
+	if (port->dev == NULL) {
+		static int next_port_number = 0;
```
chaitu236: nitpick: `static unsigned int`

chaitu236 (approved):

drivers/tty/serial/serial_core.c
sjasonsmith: If you change this be sure to change the format string to %u.

I don't really think it matters either way. 2^31 and 2^32 possibilities should both be plenty.

sjasonsmith (approved):

drivers/tty/serial/serial_core.c
sjasonsmith: I do agree that unsigned makes more sense here, despite not mattering in practice.

drivers/tty/serial/serial_core.c
```diff
@@ -3440,6 +3457,10 @@ int serial_core_register_port(struct uart_driver *drv, struct uart_port *port)
 err_unregister_ctrl_dev:
 	serial_base_ctrl_device_remove(new_ctrl_dev);
 
+	/* Delete the anon platform_device */
```
gratian: Nitpick: this comment doesn't add any information which is not already obvious from the code. I would remove it (or expand it into something more useful if you so desire).

For more details, see the "8) Commenting" section of: https://www.kernel.org/doc/html/latest/process/coding-style.html

drivers/tty/serial/serial_core.c
```diff
@@ -3470,6 +3491,10 @@ void serial_core_unregister_port(struct uart_driver *drv, struct uart_port *port
 	if (!serial_core_ctrl_find(drv, phys_dev, ctrl_id))
 		serial_base_ctrl_device_remove(ctrl_dev);
 
+	/* Delete the anon platform_device */
```
gratian: Same nitpick as above, the comment is not very useful, remove.

drivers/tty/serial/serial_core.c
gratian: Also checkpatch complains about:
```
ERROR: do not initialise statics to 0
#31: FILE: drivers/tty/serial/serial_core.c:3409:
+               static int next_port_number = 0;
```
That's because static variables already default to zero and is more efficient if you leave the initialization out (the compiler can place them in the .bss segment which gets zero-out in one go). You can safely remove the '=0' initialization.

gratian (approved):
Looks good. Just some small nitpicks. Should be ready to pull after those are addressed.

drivers/tty/serial/serial_core.c
nikhalim: Updated to `static unsigned int`

drivers/tty/serial/serial_core.c
nikhalim: Comment removed

drivers/tty/serial/serial_core.c
nikhalim: Comment removed

nikhalim:
checkpatch on the latest update
```
./scripts/checkpatch.pl 0001-serial-core-create-anonymous-parent-for-NI-987x.patch
total: 0 errors, 0 warnings, 80 lines checked

0001-serial-core-create-anonymous-parent-for-NI-987x.patch has no obvious style problems and is ready for submission.
```

PR #197, opened 2025-02-13
Revert "HACK sd: Prevent SD from doing high-speed timing"

erickshepherdNI:
This reverts commit 97ebd2d4c626a0952daf25d5d9a7c5e96b6a30f4.

### Justification
The commit being reverted only addressed an issue with the 9607 and 9627, which are both Zynq-7000 based. These devices are still using the 4.14 kernel so the change should be safe to drop from 6.6. The alternative would be to upstream the change, but would likely require some discussion on how to re-implement it since it was initially implemented as a quick hack. More details about the original change can be found [here](https://review-board.natinst.com/r/99036/#comment487189).

[AB#2316653](https://dev.azure.com/ni/DevCentral/_workitems/edit/2316653)

### DIFF
--- a/drivers/mmc/host/sdhci-pltfm.c
+++ b/drivers/mmc/host/sdhci-pltfm.c
@@ -93,9 +93,6 @@ void sdhci_get_property(struct platform_device *pdev)
 	if (device_property_present(dev, "broken-cd"))
 		host->quirks |= SDHCI_QUIRK_BROKEN_CARD_DETECTION;
 
-	if (device_property_present(dev, "force-sd-standard"))
-		host->quirks |= SDHCI_QUIRK_NO_HISPD_BIT;
-
 	if (device_property_present(dev, "no-1-8-v"))
 		host->quirks2 |= SDHCI_QUIRK2_NO_1_8_V;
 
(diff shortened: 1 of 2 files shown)


chaitu236:
There have been recent PRs in this repo to update kernel for bluefin devices (which are zynq based from what I understand) - e.g., https://github.com/ni/linux/pull/195, https://github.com/ni/linux/pull/192, etc. They may not care about 9607 and 9627 specifically but these changes may still be needed for the bluefin devices. You may want to check with them.

gratian:
@chaitu236 searching the device trees for `force-sd-standard` I only see it used in:
```
$ git grep "force-sd-standard"
arch/arm/boot/dts/xilinx/ni-sbrio-9607.dts:     force-sd-standard;
arch/arm/boot/dts/xilinx/ni-sbrio-9627.dts:     force-sd-standard;
drivers/mmc/host/sdhci-pltfm.c: if (device_property_present(dev, "force-sd-standard"))
```
This is true on both 6.6 and 4.14 branches. Considering this, I intend to merge this revert.

If we do need it in the future for some bluefin device let's talk about adding it back in a non-hacky way. FYI @HatsyRei @DeborahOoi96

gratian (approved):

erickshepherdNI:
> There have been recent PRs in this repo to update kernel for bluefin devices (which are zynq based from what I understand) - e.g., #195, #192, etc. They may not care about 9607 and 9627 specifically but these changes may still be needed for the bluefin devices. You may want to check with them.

The author of these pull requests confirmed that the bluefin devices should not need this fix.

PR #203, opened 2025-02-25
Add config flags for NAND for ni-bluefin devices

DeborahOoi96:
We want to get pl35x-nand-controller.c run at device boot up. These changes will trigger pl35x-nand-controller to be run and therefore cause NAND related messages to be displayed in PUTTY. This is the first step to updating the NAND for bluefin in 6.6 kernel.


gratian:
Can you make these defconfig changes using the savedefconfig workflow? You shouldn't need a separate commit for the round-trip through savedefconfig.

Split the bluefin and zynq changes into separate commits. While the changes are the same for the two - they are independent changes. Also the commit message:
```
Add NAND to bluefin and zynq defconfig files

Signed-off-by: deooi <deborah.ooi@ni.com>
```
is incorrect. You are not adding defconfig files - you are enabling the PL35X driver so it should be something like:

```
nati_bluefin_defconfig: enable pl35x nand driver

Signed-off-by: ...
````
and 
```
nati_zynq_defconfig: enable pl35x nand driver
...
```

DeborahOoi96:
Thanks for your comment. I will close this PR and open another one according to your suggestions.

chaitu236:
> Thanks for your comment. I will close this PR and open another one according to your suggestions.

We usually prefer force-pushing new changes to the same PR so we can keep comment/conversation history, etc in one place. I see that you've already created a new PR so it's ok for now.

PR #209, opened 2025-03-04
mtd: Create gluebi mtdblocks if CONFIG_FTL is not enabled

HatsyRei:
WI: [AB#3024196](https://ni.visualstudio.com/DevCentral/_workitems/edit/3024196)

Bluefin devices expect to mount mtdblock devices created by gluebi at runtime, but https://github.com/ni/linux/commit/cfd7c9d260dc0a3baaea05a122a19ab91e193c65 prevents gluebi from creating these mtdblock devices as a mitigation to a NULL pointer dereference caused by FTL notifier. As a workaround, this commit adds a check for `CONFIG_FTL` before doing so, keeping the solution intact if `CONFIG_FTL` is enabled, while allowing the creation of gluebi mtdblock devices if `CONFIG_FTL` is not enabled. 

Testing: Built and deployed kernel with changes on cDAQ-9189, verified that mtdblocks are enumerated for gluebi mtd partitions when `CONFIG_FTL` is not enabled.

### DIFF
--- a/drivers/mtd/mtd_blkdevs.c
+++ b/drivers/mtd/mtd_blkdevs.c
@@ -463,7 +463,8 @@ static void blktrans_notify_add(struct mtd_info *mtd)
 {
 	struct mtd_blktrans_ops *tr;
 
-	if (mtd->type == MTD_ABSENT || mtd->type == MTD_UBIVOLUME)
+	if (mtd->type == MTD_ABSENT ||
+	   (IS_ENABLED(CONFIG_FTL) && mtd->type == MTD_UBIVOLUME))
 		return;
 
 	list_for_each_entry(tr, &blktrans_majors, list)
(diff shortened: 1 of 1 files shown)


chaitu236 (commented):
You mention that this is a workaround in your PR description. What would the correct fix look like?

HatsyRei:
> You mention that this is a workaround in your PR description. What would the correct fix look like?

The other option would involve leaving the driver code as is, and utilizing a different way to emulate block devices on UBI (such as ubiblock). This would require us to pull in a new driver, make changes to u-boot to ensure that the block devices are correctly configured at boot time, and have the kernel anticipate the new block device labels and mount them. Theoretically all these should just work, but we'll need to look at whether these changes can be delivered to shipped hardware via existing field firmware update infrastructure. My understanding is we don't have an official way to update u-boot on Bluefin devices, so this will be difficult. 

The workaround here is hence to just allow enumeration of block devices for gluebi mtd partitions if the ftl driver is not enabled.

chaitu236 (approved):

gratian (approved):
The change looks fine (given the options we have) but can you mention the upstream commit directly in the commit message and not the GitHub PR description.

Somebody looking at this commit in the future (for example when doing a kernel rebase/upgrade) won't be able to easily tie it back to the PR description and understand what is going on.

HatsyRei:
Edited commit message and rebased fork.

gratian:
I've reworded the commit message slightly to match the upstream standard for quoting commits/commit messages.
Merged outside GitHub workflow; closing PR.

PR #210, opened 2025-03-06
pl35x-nand-controller: Enable dynamically setting clk rate in NAND.

DeborahOoi96:
NI Zynq based devices require different nand memclk frequencies to be set at different timing modes, as seen in the bluefin branch of [ni-zynq.dtsi](https://github.com/ni/linux/blob/2dc2aab3edd7c68cc7f96955a7763e618725c03e/arch/arm/boot/dts/ni-zynq.dtsi#L166).
 This commit enables the memclk rate to be set according to the specifications in the device tree. If the device tree does not have this memclk-timing-frequency property specified, the driver proceeds as usual.
 
WI:  [AB#3024184](https://dev.azure.com/ni/DevCentral/_workitems/edit/3024184)

 Testing: 
 
 Tested my changes on cDAQ9189, NAND timing mode was able to be set successfully and NAND device was detected.
 
image

Tested with memclk-timing-frequency property not set in device tree and we get the error as expected. This printk error message is not part of the commit, just there for my own checking. Purposely not including it in the commit as it is not actually an error if the device does not have that property.
image

Tested with passing a non-existent timing mode (99) into the timing mode parameter of of_get_memclk_freq. Got the expected error.
image

### DIFF
--- a/arch/arm/boot/dts/xilinx/ni-zynq.dtsi
+++ b/arch/arm/boot/dts/xilinx/ni-zynq.dtsi
@@ -140,6 +140,16 @@
 
 &smcc {
 	status = "okay";
+	/*
+	 * NI Zynq based devices require different memclk frequency
+	 * for different timing modes
+	 */
+	memclk-timing-frequency = <0 83333333>,
+				<1 166666666>,
+				<2 166666666>,
+				<3 166666666>,
+				<4 166666666>,
+				<5 166666666>;
 };
 
 &nfc0 {
--- a/drivers/mtd/nand/raw/pl35x-nand-controller.c
+++ b/drivers/mtd/nand/raw/pl35x-nand-controller.c
@@ -795,6 +795,28 @@ static int pl35x_nfc_exec_op(struct nand_chip *chip,
 				      op, check_only);
 }
 
+static unsigned long of_get_memclk_freq(const struct device_node *np, const int timing_mode)
+{
+	int memclk_frequency, i, total_elements;
+	u32 memclk_mode;
+
+	total_elements = of_property_count_u32_elems(np, "memclk-timing-frequency");
+	if (total_elements <= 0)
+		return 0;
+
+	for (i = 0; i < total_elements; i = i+2) {
+		of_property_read_u32_index(np, "memclk-timing-frequency", i, &memclk_mode);
+		if (memclk_mode == timing_mode) {
+			of_property_read_u32_index(np, "memclk-timing-frequency",
+				i+1, &memclk_frequency);
+			return memclk_frequency;
+		}
+	}
+
+	pr_err("Failed to find matching memclk timing mode %d\n", timing_mode);
+	return 0;
+}
+
 static int pl35x_nfc_setup_interface(struct nand_chip *chip, int cs,
 				     const struct nand_interface_config *conf)
 {
@@ -804,6 +826,8 @@ static int pl35x_nfc_setup_interface(struct nand_chip *chip, int cs,
 	const struct nand_sdr_timings *sdr;
 	unsigned int period_ns, val;
 	struct clk *mclk;
+	unsigned long memclk_of_timing_freq;
+	int ret = 0;
 
 	sdr = nand_get_sdr_timings(conf);
 	if (IS_ERR(sdr))
@@ -813,6 +837,23 @@ static int pl35x_nfc_setup_interface(struct nand_chip *chip, int cs,
 	if (IS_ERR(mclk)) {
 		dev_err(nfc->dev, "Failed to retrieve SMC memclk\n");
 		return PTR_ERR(mclk);
+	}
+
+	/*
+	 * NI Zynq based devices require different rates for different timing modes
+	 * so we need to set the rate according to the timing modes
+	 */
+	memclk_of_timing_freq = of_get_memclk_freq(
+		nfc->dev->parent->of_node,
+		conf->timings.mode);
+
+	if (memclk_of_timing_freq) {
+		ret = clk_set_rate(mclk, memclk_of_timing_freq);
+		if (ret) {
+			pr_err("Failed to set memclk timing frequency to %lu\n",
+				memclk_of_timing_freq);
+			return ret;
+		}
 	}
 
 	/*


drivers/mtd/nand/raw/pl35x-nand-controller.c
```diff
@@ -795,6 +795,22 @@ static int pl35x_nfc_exec_op(struct nand_chip *chip,
 				      op, check_only);
 }
 
+static unsigned long get_special_timing_freq(const struct device_node *np, int timing_mode)
```
chaitu236: nit: `const int timing_mode` is better.

drivers/mtd/nand/raw/pl35x-nand-controller.c
```diff
@@ -815,6 +833,20 @@ static int pl35x_nfc_setup_interface(struct nand_chip *chip, int cs,
...
+	 * so we need to set the rate according to the timing modes
+	 */
+	special_timing_freq = get_special_timing_freq(
+		nfc->dev->parent->of_node,
+		conf->timings.mode);
+
+	if (special_timing_freq) {
+		ret = clk_set_rate(mclk, special_timing_freq);
+		if (ret)
+			return -EINVAL;
```
chaitu236: Can't we just return `ret`?

chaitu236 (changes requested):
image

nit: I don't think tab/spacing in commit description is required.

arch/arm/boot/dts/xilinx/ni-zynq.dtsi
```diff
@@ -140,6 +140,16 @@
...
+	/*
+	 * Bluefin devices require different memclk frequency
+	 * for different timing modes
+	 */
+	memclk-timing-frequency = <0 83333333>,
+							<1 166666666>,
+							<2 166666666>,
+							<3 166666666>,
+							<4 166666666>,
+							<5 166666666>;
```
HatsyRei: Indent misalignment most likely due to IDE set to show tab as 4 spaces. Let's stick to tab being 8 chars.
https://www.kernel.org/doc/html/latest/process/coding-style.html#indentation

chaitu236 (approved):

drivers/mtd/nand/raw/pl35x-nand-controller.c
```diff
@@ -795,6 +795,22 @@ static int pl35x_nfc_exec_op(struct nand_chip *chip,
 				      op, check_only);
 }
 
+static unsigned long get_special_timing_freq(const struct device_node *np, const int timing_mode)
```
gratian: The name of this function is too generic and it's not immediately obvious what's 'special' about it. Reading properties from the device tree is not really that special.

I would suggest renaming it to something that matches what it does, like: of_get_memclk_freq()

drivers/mtd/nand/raw/pl35x-nand-controller.c
```diff
@@ -804,6 +820,8 @@ static int pl35x_nfc_setup_interface(struct nand_chip *chip, int cs,
 	const struct nand_sdr_timings *sdr;
 	unsigned int period_ns, val;
 	struct clk *mclk;
+	unsigned long special_timing_freq;
```
gratian: Same comment as above, the name is too generic. Maybe 'memclk_of_timing_freq' or something like that.

drivers/mtd/nand/raw/pl35x-nand-controller.c
```diff
@@ -815,6 +833,22 @@ static int pl35x_nfc_setup_interface(struct nand_chip *chip, int cs,
...
+	 * so we need to set the rate according to the timing modes
+	 */
+	special_timing_freq = get_special_timing_freq(
+		nfc->dev->parent->of_node,
+		conf->timings.mode);
+
+	if (special_timing_freq) {
+		ret = clk_set_rate(mclk, special_timing_freq);
+		if (ret) {
+			pr_err("Failed to set special timing rates\n");
```
gratian: This error could use more details. Print the actual timing rate you are trying to set. And don't call it special, call it what it is (i.e. a fixed value read from the device tree).

drivers/mtd/nand/raw/pl35x-nand-controller.c
```diff
@@ -795,6 +795,22 @@ static int pl35x_nfc_exec_op(struct nand_chip *chip,
...
+{
+	int len, i;
+
+	const __be32 *memclk_timing_freq = of_get_property(np, "memclk-timing-frequency", &len);
+
+	if (memclk_timing_freq) {
+		for (i = 0; i < len / sizeof(u32) / 2; i++) {
+			if (timing_mode == be32_to_cpu(memclk_timing_freq[i * 2])) {
+				return be32_to_cpu(memclk_timing_freq[i * 2 + 1]);
+			};
```
gratian: Extra ';' and unnecessary '}'. See the section on "3) Placing Braces and Spaces" of the kernel's coding style document: https://www.kernel.org/doc/html/latest/process/coding-style.html

Specifically "Do not unnecessarily use braces where a single statement will do."

drivers/mtd/nand/raw/pl35x-nand-controller.c
```diff
@@ -795,6 +795,22 @@ static int pl35x_nfc_exec_op(struct nand_chip *chip,
...
+static unsigned long get_special_timing_freq(const struct device_node *np, const int timing_mode)
+{
+	int len, i;
+
+	const __be32 *memclk_timing_freq = of_get_property(np, "memclk-timing-frequency", &len);
+
+	if (memclk_timing_freq) {
+		for (i = 0; i < len / sizeof(u32) / 2; i++) {
+			if (timing_mode == be32_to_cpu(memclk_timing_freq[i * 2])) {
+				return be32_to_cpu(memclk_timing_freq[i * 2 + 1]);
```
gratian: In theory this could index beyond the end of the array depending on the value of 'len' and how the 'len/sizeof(u32)/2' math in the for loop gets truncated.

Not sure yet if there is a better way to iterate through this array. I'll have to do some digging through the kernel to see if there are other examples.

drivers/mtd/nand/raw/pl35x-nand-controller.c
DeborahOoi96: thanks for the name suggestion. resolved.

drivers/mtd/nand/raw/pl35x-nand-controller.c
DeborahOoi96: thanks for the name suggestion. resolved.

drivers/mtd/nand/raw/pl35x-nand-controller.c
DeborahOoi96: I found a way that involves using of_property_for_each_u32 instead. I think that is safer way to iterate through. Let me know what you think.

drivers/mtd/nand/raw/pl35x-nand-controller.c
```diff
@@ -795,6 +795,25 @@ static int pl35x_nfc_exec_op(struct nand_chip *chip,
...
 }
 
+static unsigned long of_get_memclk_freq(const struct device_node *np, const int timing_mode)
+{
+	struct property	*prop;
+	const __be32 *cur;
+	int memclk_frequency = 0;
+	u32 memclk_mode;
+
+	of_property_for_each_u32(np, "memclk-timing-frequency", prop, cur, memclk_mode) {
```
chaitu236: I've only seen `of_property_for_each_u32` being called with 3 arguments and it doesn't seem to be a variadic function either. What am I missing?

drivers/mtd/nand/raw/pl35x-nand-controller.c
```diff
@@ -795,6 +795,25 @@ static int pl35x_nfc_exec_op(struct nand_chip *chip,
...
+		if (timing_mode != memclk_mode)
+			continue;
+	
+		of_property_read_u32_index(np, "memclk-timing-frequency",
+						(cur - (__be32 *)prop->value) / sizeof(u32) + 1,
```
chaitu236: > (cur - (__be32 *)prop->value) / sizeof(u32) + 1, 

This also looks weird as well. https://elixir.bootlin.com/linux/v6.6.82/source/drivers/irqchip/irq-mvebu-gicp.c#L187 this seems like a better way to achieve what you're trying to do. You may want to search in kernel to figure out the best way to do this.

chaitu236 (changes requested):

drivers/mtd/nand/raw/pl35x-nand-controller.c
DeborahOoi96: Thanks for the suggestion, I have modified it to use the new way.

drivers/mtd/nand/raw/pl35x-nand-controller.c
```diff
@@ -795,6 +795,28 @@ static int pl35x_nfc_exec_op(struct nand_chip *chip,
 				      op, check_only);
 }
 
+static unsigned long of_get_memclk_freq(const struct device_node *np, const int timing_mode)
+{
+	int memclk_frequency, i, total_elements = 0;
```
chaitu236: nit: initializing `total_elements` is unnecessary right?

drivers/mtd/nand/raw/pl35x-nand-controller.c
```diff
@@ -795,6 +795,28 @@ static int pl35x_nfc_exec_op(struct nand_chip *chip,
...
+	for (i = 0; i < total_elements; i = i+2) {
+		of_property_read_u32_index(np, "memclk-timing-frequency", i, &memclk_mode);
+		if (memclk_mode == timing_mode) {
+			of_property_read_u32_index(np, "memclk-timing-frequency",
+				i+1, &memclk_frequency);
+			return memclk_frequency;
+		}
+	}
+
+	pr_err("Failed to find matching timing mode %d\n", timing_mode);
```
chaitu236: Can you make the error string more specific? E.g., mention memclk like you've done below.

chaitu236 (changes requested):
Your PR description only mentions testing for "happy path".

Did you also test all the possible failure paths in this code?

drivers/mtd/nand/raw/pl35x-nand-controller.c
chaitu236: @DeborahOoi96 I know this code isn't relevant anymore but I'd still like to understand how this worked.

DeborahOoi96:
> Your PR description only mentions testing for "happy path".
> 
> Did you also test all the possible failure paths in this code?

Yes I did. Updated the PR description with all the test cases and their outcomes.

arch/arm/boot/dts/xilinx/ni-zynq.dtsi
```diff
@@ -140,6 +140,16 @@
 
 &smcc {
 	status = "okay";
+	/*
+	 * Bluefin devices require different memclk frequency
```
gratian: This comment and the commit in general implies this applies only to Bluefin devices. However the way you have it implemented right now it will affect all NI Zynq based devices (since this is a common include file). Not necessarily saying that's wrong (I could see this being needed for other Zynq based designs), just that the comments/commit doesn't match what the code does.

If the intent is to apply this to all NI Zynq devices I would modify the commit message and the code comments to reflect that (ie. use "NI Zynq based devices" in places where you currently have "Bluefin"). Otherwise I guess this needs to be moved into the Bluefin specific device tree. Sorry I didn't notice this earlier.

DeborahOoi96: If I recall correctly, this particular file is not included anywhere else in 6.6 branch besides the bluefin device family tree. Only ni-bluefin.dts, ni-bluefinlc.dts, ni-coralreef.dts, and ni-fielddaq.dts include this file, and all of them do need this specific timing mode. Keeping it here will reduce code duplication, but I see how it can be confusing.

What would you suggest the comment be in this case? "NI Zynq based devices" might imply a scope larger than the bluefin family tree.

gratian: At the moment, yes, only bluefin devices are included in the 6.6 branch but I'm not sure that will be true in the future. i.e. if we and another Zynq variant to the 6.6 branch this will apply to it automatically so I don't think "NI Zynq based devices" is too broad in scope - It actually reflects the true scope of the change.

We can always change the scope later if we have to (by creating a bluefin specific include) but as it stands I want the commit/comments to accurately express the current scope it applies to.

DeborahOoi96: Updated code comments and commit message to "NI Zynq based devices" instead of bluefin where applicable.

gratian (approved):

chaitu236 (approved):

PR #212, opened 2025-03-12
serial: core: add checking for platform device

nikhalim:
Check whether parent device is of the anon_device_type before getting its platform_device

### DIFF
--- a/drivers/tty/serial/serial_core.c
+++ b/drivers/tty/serial/serial_core.c
@@ -46,6 +46,10 @@ static DEFINE_MUTEX(port_mutex);
  */
 static struct lock_class_key port_lock_key;
 
+static struct device_type anon_dev_type = {
+	.name = "anon_port",
+};
+
 #define HIGH_BITS_OFFSET	((sizeof(long)-sizeof(int))*8)
 
 /*
@@ -3396,6 +3400,7 @@ int serial_core_register_port(struct uart_driver *drv, struct uart_port *port)
 {
 	struct serial_ctrl_device *ctrl_dev, *new_ctrl_dev = NULL;
 	int ret;
+	bool is_parent_anon_port = false;
 
 	mutex_lock(&port_mutex);
 
@@ -3419,6 +3424,8 @@ int serial_core_register_port(struct uart_driver *drv, struct uart_port *port)
 		}
 
 		port->dev = &pdev->dev;
+		port->dev->type = &anon_dev_type;
+		is_parent_anon_port = true;
 	}
 
 	/* Inititalize a serial core controller device if needed */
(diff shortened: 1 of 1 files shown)


nikhalim:
@sjasonsmith @gratian Please help to review this PR

drivers/tty/serial/serial_core.c
```diff
@@ -3476,6 +3483,7 @@ void serial_core_unregister_port(struct uart_driver *drv, struct uart_port *port
 	struct serial_port_device *port_dev = port->port_dev;
 	struct serial_ctrl_device *ctrl_dev = serial_core_get_ctrl_dev(port_dev);
 	int ctrl_id = port->ctrl_id;
+	struct platform_device *pdev;
```
sjasonsmith: You are not using this.

sjasonsmith (commented):
Have you tested this? Ideally you would test both that it fixes the crashing you are seeing when pciserial unloads, as well as that it continues to work through surprise removal sequences with cRIO.

sjasonsmith:
@gratian this technique was derived from this old post where a similar issue was resolved elsewhere.
https://fa.linux.kernel.narkive.com/3rmBMJOU/patch-v2-mfd-only-unregister-platform-devices-allocated-by-the-mfd-core

drivers/tty/serial/serial_core.c
```diff
@@ -3457,7 +3464,7 @@ int serial_core_register_port(struct uart_driver *drv, struct uart_port *port)
 err_unregister_ctrl_dev:
 	serial_base_ctrl_device_remove(new_ctrl_dev);
 
-	if (strncmp(to_platform_device(port->dev)->name, "anon_port", 9) == 0)
+	if (is_parent_null)
```
chaitu236: nitpick: Seeing that if the parent is not NULL anymore after `port->dev = &pdev->dev;` above, naming this `is_parent_anon_port` or something like that might be better.

chaitu236 (changes requested):

gratian (commented):
In general it looks good. In addition to the improvements suggested @sjasonsmith, and @chaitu236 the commit message could use some tweaking so that in the future we know why the change was needed. It would also let us know that it should be combined with the commit it fixes on the next rebase (it also needs a 'Fixes' tag).

I would reword it to say something like:

```
Commit 978ca75850f2 ("serial: core: create anonymous parent device")
introduced a NULL pointer dereference by accessing ...

Check whether parent device is of the anon_device_type before getting
its platform_device.

Fixes: 978ca75850f2 ("serial: core: create anonymous parent device")
Signed-off-by: <your signoff>
```

drivers/tty/serial/serial_core.c
nikhalim: Fixed.

drivers/tty/serial/serial_core.c
nikhalim: Fixed.

nikhalim:
All feedbacks addressed.

```
./scripts/checkpatch.pl 0001-serial-core-add-checking-for-platform-device.patch
total: 0 errors, 0 warnings, 41 lines checked

0001-serial-core-add-checking-for-platform-device.patch has no obvious style problems and is ready for submission.
```

gratian (approved):

PR #213, opened 2025-03-12
mmc: Add quirk to disable DDR50 tuning

erickshepherdNI:
Adds the MMC_QUIRK_NO_UHS_DDR50_TUNING quirk and updates
mmc_execute_tuning() to return 0 if that quirk is set. This fixes an
issue on certain Swissbit SD cards that do not support DDR50 tuning
where tuning requests caused I/O errors to be thrown.

[WI#2805437](https://dev.azure.com/ni/DevCentral/_workitems/edit/2805437)

# Testing
Built the kernel and confirmed that the SD card attempts to tune but is stopped by the new change. No I/O errors were thrown.


drivers/mmc/core/quirks.h
```diff
@@ -34,6 +34,9 @@ static const struct mmc_fixup __maybe_unused mmc_sd_fixups[] = {
 		   MMC_QUIRK_BROKEN_SD_CACHE | MMC_QUIRK_BROKEN_SD_POWEROFF_NOTIFY,
 		   EXT_CSD_REV_ANY),
 
+	MMC_FIXUP("0016G", CID_MANFID_SWISSBIT, 0x5342, add_quirk,
```
chaitu236: Like the other fixups in this struct, can you add a comment describing why this fixup is needed.

chaitu236 (changes requested):
Changes lgtm. But please add WI# to PR description and rebase to latest `nilrt/master/6.6`.

Btw, did you end up testing on other controllers? What were the findings?

drivers/mmc/core/quirks.h
erickshepherdNI: Ok, I added a comment here.

erickshepherdNI:
> Changes lgtm. But please add WI# to PR description and rebase to latest `nilrt/master/6.6`.
> 
> Btw, did you end up testing on other controllers? What were the findings?

I updated the PR description and rebased the changes. as far as other testing, I was not able to reproduce the issue on a 9034 with the same Swissbit card. I'd be curious to test on more hardware configurations but this change should be ok since the card doesn't support tuning anyway.

chaitu236:
> > Changes lgtm. But please add WI# to PR description and rebase to latest `nilrt/master/6.6`.
> > Btw, did you end up testing on other controllers? What were the findings?
> 
> I updated the PR description and rebased the changes. as far as other testing, I was not able to reproduce the issue on a 9034 with the same Swissbit card. I'd be curious to test on more hardware configurations but this change should be ok since the card doesn't support tuning anyway.

I see. Did we get the info about it's tuning support from it's manual/manufacturer?

erickshepherdNI:
> > > Changes lgtm. But please add WI# to PR description and rebase to latest `nilrt/master/6.6`.
> > > Btw, did you end up testing on other controllers? What were the findings?
> > 
> > 
> > I updated the PR description and rebased the changes. as far as other testing, I was not able to reproduce the issue on a 9034 with the same Swissbit card. I'd be curious to test on more hardware configurations but this change should be ok since the card doesn't support tuning anyway.
> 
> I see. Did we get the info about it's tuning support from it's manual/manufacturer?

I dug up the exact card model and spec sheet when I started working on this and confirmed that the card was built to the 3.0 SD spec that does not support DDR50 tuning. I can find the documentation again and attach it to the work item. It would probably be good to have for future reference.

chaitu236 (approved):

drivers/mmc/core/quirks.h
```diff
@@ -34,6 +34,14 @@ static const struct mmc_fixup __maybe_unused mmc_sd_fixups[] = {
 		   MMC_QUIRK_BROKEN_SD_CACHE | MMC_QUIRK_BROKEN_SD_POWEROFF_NOTIFY,
 		   EXT_CSD_REV_ANY),
 
+	/*
+	 * Certain DDR50 Swissbit SD cards built to the 3.0 SD spec, which
```
gratian: I suspect the feedback you will get upstream on this comment is that is too vague. I see the other comments in this struct try to add details like manufacturing dates. Can we provide more details here like the specific Siwssbit card model and/or manufacturing date?

gratian (approved):
Looks good. Just a nitpick on the comment being too general to be useful.

Also looks like check-patch flagged an indentation warning (see the 'Checks' tab in the PR).

drivers/mmc/core/quirks.h
erickshepherdNI: I can add that information. Should I also specify the manufacturing date when applying the quirk? I wasn't sure if that date would be the same for all cards of this model.

chaitu236:
@erickshepherdNI can this be closed?

PR #214, opened 2025-03-12
mmc: host: Wait for Vdd to settle on card power off

erickshepherdNI:
The SD spec version 6.0 section 6.4.1.5 requires that Vdd must be
lowered to less than 0.5V for a minimum of 1 ms when powering off a
card. Increase wait to 15 ms so that voltage has time to drain down
to 0.5V and cards can power off correctly. Issues with voltage drain
time were only observed on Apollo Lake and Bay Trail host controllers
so this fix is limited to those devices.

[WI#2316635](https://dev.azure.com/ni/DevCentral/_workitems/edit/2316635)

# Testing
Built the kernel locally and confirmed the 15ms delay was present. Tested on a cRIO-9053 and cRIO-9043 to cover Apollo Lake and Bay Trail host controllers.

### DIFF
--- a/drivers/mmc/host/sdhci-pci-core.c
+++ b/drivers/mmc/host/sdhci-pci-core.c
@@ -608,8 +608,12 @@ static void sdhci_intel_set_power(struct sdhci_host *host, unsigned char mode,
 
 	sdhci_set_power(host, mode, vdd);
 
-	if (mode == MMC_POWER_OFF)
+	if (mode == MMC_POWER_OFF) {
+		if (slot->chip->pdev->device == PCI_DEVICE_ID_INTEL_APL_SD ||
+		    slot->chip->pdev->device == PCI_DEVICE_ID_INTEL_BYT_SD)
+			usleep_range(15000, 17500);
 		return;
+	}
 
 	/*
 	 * Bus power might not enable after D3 -> D0 transition due to the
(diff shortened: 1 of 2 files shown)


erickshepherdNI:
Adrian, the upstream maintainer, suggested that I move the delay to the set_power() callback instead of set_ios(). sdhci_set_power() gets called right before the delay in sdhci_set_ios() so I would expect this behavior to be the same. Adrian also suggested I use the usleep_range() function instead of mdelay() but didn't specify exactly why. He also wanted me to make the delay specific to Apollo Lake devices but I asked Tony Liechty, who was involved with this issue initially, and he believes the delay should be used by all of our devices. I'm going to get Adrian's opinion on this before pushing more changes upstream.

chaitu236:
> Adrian also suggested I use the usleep_range() function instead of mdelay() but didn't specify exactly why

Possibly because `mdelay` wastes cpu cycles by busy spinning and is usually used in interrupt contexts where *sleep isn't allowed. `msleep`/`usleep*` lets scheduler schedule out the thread and run something else until the sleep completes and thread is runnable again.

gratian (approved):

chaitu236 (changes requested):
Changes look fine but the commit message should be made more generic around "Increase our wait..". Remove "our" and mention why 15ms is required and not 1ms as the spec says - i.e., some controllers require longer waits, etc.

chaitu236 (approved):

chaitu236:
@erickshepherdNI can this be closed?

PR #188, opened 2025-01-14
devicetree, ni-bluefin: Cherry-pick all device tree commits related to bluefin

DeborahOoi96:
We need to update linux 6.6 kernel to support bluefin devices. Cherry-picking all  device tree commits related to bluefin from bluefin/master/4.1 into nilrt/master/6.6. These changes are needed in order for linux-bluefin kernel recipe to build correctly.

WI: [AB#2959873](https://dev.azure.com/ni/DevCentral/_workitems/edit/2959873)

Testing
Unable to build kernel until all necessary commits are cherry-picked.

PR #216, opened 2025-03-18
Enable Marvell ethernet switch for Bluefin devices

HatsyRei:
WI: [AB#3033955](https://ni.visualstudio.com/DevCentral/_workitems/edit/3033955)

Bluefin ethernet interface is exposed via a Marvell 88E6341 'Topaz' ethernet switch. These commits pull in the mv88e6xxx driver by enabling `CONFIG_NET_DSA_MV88E6XXX` and its PTP support via `CONFIG_NET_DSA_MV88E6XXX_PTP`. Device tree bindings are added to Bluefin device tree to enable probe and configuration of the switch at runtime. 

Testing: Booted kernel with changes on cDAQ-9189, verified functionality of ethernet ports, sanity check output of ifconfig and content of /sys/class/ptp.


arch/arm/configs/nati_bluefin_defconfig
```diff
@@ -60,7 +60,6 @@ CONFIG_BRIDGE_EBT_MARK_T=y
 CONFIG_BRIDGE=y
 CONFIG_BRIDGE_VLAN_FILTERING=y
 CONFIG_NET_DSA=y
-CONFIG_NET_DSA_MV88E6XXX_ONLY_8021AS=y
```
chaitu236, line 63: I assume this got dropped because of `savedefconfig`? Does that mean you don't require the https://github.com/ni/linux/commit/295c911ba097730c1f3fc0974145b50bc5e414bc cherrypick? I was wondering if you needed that when doing 6.12 [rebase](https://github.com/ni/linux/pull/211).

HatsyRei: Yes, these are the results of `savedefconfig`. My understanding is that the 6.6 branch already has upstream DSA changes which obviates the need for https://github.com/ni/linux/commit/295c911ba097730c1f3fc0974145b50bc5e414bc, and it got included in https://github.com/ni/linux/pull/188 due to an oversight. Should I revert this particular commit?

arch/arm/boot/dts/xilinx/ni-bluefin.dts
```diff
@@ -75,6 +72,55 @@
...
+		status = "okay";
+
+		ports {
+			#address-cells = <1>;
+			#size-cells = <0>;
+
+			port@0 {
+				reg = <0>;
+				label = "cpu";
+				phy-mode = "rgmii-id";
```
HatsyRei, line 97: Fixed an issue with phy-mode specification. Now using "rgmii-id" (rgmii with internal delays) as that appears to be the correct configuration for switch on Bluefin, based on https://github.com/ni/linux/commit/8593a737dd13f343e7e4cb8f51919d5cee94bd27. Verified network connectivity after deploying changes on 9189.

arch/arm/configs/nati_bluefin_defconfig
chaitu236: Ok, I think this is fine.

chaitu236 (approved):

gratian (approved):

PR #230, opened 2025-04-23
nati_bluefin_defconfig: Enable uevent helper

HatsyRei:
WI: [AB#3077142](https://ni.visualstudio.com/DevCentral/_workitems/edit/3077142)

Encountered some issues when reserving Bluefin as a network devices on MAX when running Linux 6.6. Turns out the hotplug script responsible for mounting Umber and chowning its corresponding dev entry did not execute, resulting in daqbridge failing to communicate with the Umber device.

Testing: Proper execution of the hotplug event handler script is observed after specifying its path in Bluefin's defconfig. Timeout issues with reserving network device no longer reproducible. 

P.S. Some further info:
- In 4.1, the hotplug script works because the path was specified within sysctl.conf, which to my understanding is consumed if [CONFIG_SYSCTL_SYSCALL](https://github.com/ni/linux/blob/2dc2aab3edd7c68cc7f96955a7763e618725c03e/arch/arm/configs/nati_bluefin_defconfig#L17) is enabled in defconfig.
- CONFIG_SYSCTL_SYSCALL has been deprecated and removed from Linux kernel since https://github.com/ni/linux/commit/61a47c1ad3a4dc6882f01ebdc88138ac62d0df03.


chaitu236 (approved):

gratian (approved):

gratian:
Cherry picked in `nilrt/master/6.12` as commit https://github.com/ni/linux/commit/d16fd4ccd69e344a7407601ae19f158495ccad12

PR #234, opened 2025-05-06
nati_x86_64_defconfig: enable all kernel modules

rajendra-desai-ni:
Enabled all the loadable modules in the default defconfig file so that they are available for installation.
Used ubuntu linux kernel config file as a reference to enable all the loadable modules in the kernel.
Also regenerated nati_x86_64_defconfig file.

[AB#2536400](https://dev.azure.com/ni/DevCentral/_workitems/edit/2536400/)

### Testing
- [x] Built the kernel image and the modules with existing changes
- [x] Installed the kernel image and the modules on a cRIO-9030 target
- [x] Verified the installation of modules using modprobe and insmod

### Note to reviewers
There will be a sustaining WI under the kernel upstream feature to keep track of periodic updates to the defconfig file.


chaitu236 (changes requested):
These new modules will be present in feeds but not pre-installed on the BSI, correct?

Also, check patch failing

image

rajendra-desai-ni:
> These new modules will be present in feeds but not pre-installed on the BSI, correct?

Correct

> Also, check patch failing
> 
> image

I have regenerated the defconfig file and this step is not failing anymore

gratian:
We have decided to use a different approach. See: https://github.com/ni/meta-nilrt/pull/834

PR #260, opened 2025-10-20
mv88e6xxx fixes for 1588 with trunked ports

bstreiff:
Two patches here, technically they're separate things but related.

- embed rx timestamps into packets
    - our 4.1 pre-upstream implementation of mv88e6xxx hardware timestamping used a feature on the switch to embed the hardware's 32-bit timestamp counter into the "reserved2" field of an incoming PTP packet, but the implementation that landed upstream uses an older method of reading out the timestamp counter over SMI for the sake of older generations of hardware. I'm seeing a large number of timestamp overruns with this approach though in our test networks, though, especially in instances where both layer 2 (802.1AS) and layer 4 (1588 UDP) masters are running.
    - fixes [AB#3396801](https://dev.azure.com/ni/DevCentral/_workitems/edit/3396801/)
    -  [Upstream mailing list traffic suggests that there are PTP testers that will also trigger this](https://lore.kernel.org/netdev/DD8PM8593EAM.1FQU37SU51BLJ@gmail.com/); but also there's a big outstanding patchset that refactors Marvell timestamping to share code amongst the PHYs and switch products (see that whole thread), so this may need to be reworked with that patchset in mind to bring it upstream.
    - an upstream quibble I would expect is the use of a switch statement for checking families rather than extending `struct mv88e6xxx_ptp_ops` to indicate whether the switch chip has this capability.
- add 1588-related addresses to ATU as management frames
    - when ports are trunked due to being placed in a link-aggregated bond, 1588 packets weren't getting delivered to the port interfaces anymore; this was because the packets were getting a DSA tag indicating that they're part of a trunk, and [the DSA infrastructure applies them to the bond interface in that case](https://github.com/ni/linux/blob/9f2f3965b02c845995c14b390e549fc7dc0bfc5d/net/dsa/tag_dsa.c#L263). The solution is to program the ATU to indicate that packets with 1588-related destination addresses are to be treated as management frames and trapped to the CPU, which gives them the same behavior that the "rsvd2cpu" function already applies for `01:80:c2:00:00:0x`. We do this for every VLAN, regardless of whether or not hardware timestamping is enabled.
    - required for [AB#3432087](https://dev.azure.com/ni/DevCentral/_workitems/edit/3432087/)

### DIFF
--- a/drivers/net/dsa/mv88e6xxx/chip.c
+++ b/drivers/net/dsa/mv88e6xxx/chip.c
@@ -2515,6 +2515,64 @@ static int mv88e6xxx_port_broadcast_sync(struct mv88e6xxx_chip *chip, int port,
 				  &ctx);
 }
 
+/* The rsvd2cpu function can trap frames with destination addresses of
+ * 01:80:c2:00:00:0x and 01:80:c2:00:00:2x as management, but there are
+ * other such frames that need to be trapped to the CPU, such as destination
+ * addresses associated with 1588, that are not in those ranges.
+ */
+
+static const u8 mv88e6xxx_extra_mgmt_da[][ETH_ALEN] = {
+	/* 1588 layer 2 */
+	{ 0x01, 0x1b, 0x19, 0x00, 0x00, 0x00 },
+	/* 1588 IPv4: 224.0.1.129 */
+	{ 0x01, 0x00, 0x5e, 0x00, 0x01, 0x81 },
+	/* 1588 IPv4 P2P: 224.0.0.107 */
+	{ 0x01, 0x00, 0x5e, 0x00, 0x00, 0x6b },
+	/* 1588 IPv6: ff0e::181 */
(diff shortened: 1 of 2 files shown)


gratian (approved):
@bstreiff Do you need this pulled into `nilrt/master/6.6`? As of last quarter NILRT moved to building and shipping `nilrt/master/6.12` so 6.6 is not getting built or being updated anymore.

bstreiff:
[linux-bluefin](https://github.com/ni/meta-expio/blob/nilrt/master/scarthgap/meta-expio/recipes-kernel/linux/linux-bluefin_git.bb) is currently building against nilrt/master/6.6, but I'm waiting to hear back on if that should be bumped to 6.12. (At least in a cursory spot-check a build with 6.12 seems to be working without issue.)

bstreiff:
Closing in favor of targeting 6.12 instead: #261

PR #266, opened 2026-01-22
nati_x86_64_defconfig: switch kernel compression to xz

gratian:
Switch the kernel compression algorithm to xz from previous gzip default.

In testing, cRIO-903x and cRIO-904x show a 2.5MB (27.2%) kernel image size decrease.

This also improves the boot time by
   - 4.8 seconds (12.8%) on cRIO-903x
   - 2.9 seconds (10%) on cRIO-9043


chaitu236 (approved):

chaitu236:
So does the kernel uncompress itself during boot? And we don't need any changes in bootloader?

gratian:
Yes, the kernel uncompresses itself, no changes needed to the bootloader.

PR #271, opened 2026-03-25
dts: Add ni-bluefinlc device tree

DeborahOoi96:
Device tree for Bluefin LC devices are for Low-cost Ethernet cDAQ-9183 and cDAQ-9187.
The changes here is to update the model name, NIDEVCODE and removal of dsa switch port 2 (sw1) since Bluefin LC devices only have a single ethernet port. Otherwise, the rest of the device tree remains the same as Bluefin.

Testing:
When built together with these changes https://github.com/ni/meta-expio/pull/51, we were able to get ni-bluefinlc.dtb generated.
image

### DIFF
--- a/arch/arm/boot/dts/xilinx/ni-bluefinlc.dts
+++ b/arch/arm/boot/dts/xilinx/ni-bluefinlc.dts
@@ -1,0 +1,138 @@
+// SPDX-License-Identifier: GPL-2.0
+/dts-v1/;
+#include <dt-bindings/interrupt-controller/irq.h>
+/include/ "ni-zynq.dtsi"
+
+/* NIDEVCODE 7B37 */
+/* NIDEVCODE 7B38 */
+
+/ {
+	model = "NI Bluefin LC";
+	compatible = "ni,zynq", "xlnx,zynq-7000";
+
+	leds {
+		compatible = "gpio-leds";
+
+		/* LED_STATUSy on GPIO46 */
+		status {
+			label = "nilrt:status:yellow";
+			gpios = <&gpio0 46 0>;
+			default-state = "on";
+		};
+
+		/* LED_ACTIVEg on GPIO47 */
+		active {
+			label = "nilrt:active:green";
+			gpios = <&gpio0 47 0>;
+			default-state = "off";
+		};
+	};
+
+	gpio_keys {
+		compatible = "gpio-keys";
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		/* Reset switch is on GPIO48 */
+		reset_sw@0 {
+			label = "reset_sw";
+			gpios = <&gpio0 48 1 /* GPIO_ACTIVE_LOW */>;
+			linux,code = <408>; /* KEY_RESTART */
+			wakeup-source;
+		};
+	};
+
+	gpio_restart {
+		compatible = "gpio-restart";
+
+		/* ~PS_FORCE_RESET is on GPIO44 */
+		gpios = <&gpio0 44 1 /* GPIO_ACTIVE_LOW */>;
+		priority = <200>;
+	};
+};
+
+&i2c0 {
+	/* Override ni-zynq.dtsi; we do not have a CPLD at 0x40. */
+	nicpld@40 {
+		status = "disabled";
+	};
+
+	tmp451@4c {
+		compatible = "ti,tmp451";
+		reg = <0x4c>;
+		vcc-supply = <&regulator_vccpint>;
+	};
+};
+
+&gem0 {
+	status = "okay";
+	#address-cells = <1>;
+	#size-cells = <0>;
+
+	fixed-link {
+		speed = <1000>;
+		full-duplex;
+		reg = <0>;
+	};
+
+	mdio0: mdio {
+		status = "okay";
+	};
+};
+
+&mdio0 {
+	#address-cells = <1>;
+	#size-cells = <0>;
+
+	switch: switch@0 {
+		compatible = "marvell,mv88e6085";
+		reg = <0>;
+		status = "okay";
+		interrupt-parent = <&intc>;
+		interrupts = <45 IRQ_TYPE_LEVEL_LOW>;
+		interrupt-controller;
+		#interrupt-cells = <2>;
+
+		ports {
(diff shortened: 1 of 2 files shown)


arch/arm/boot/dts/xilinx/ni-bluefinlc.dts
```diff
@@ -0,0 +1,138 @@
+// SPDX-License-Identifier: GPL-2.0
```
chaitu236, line 1: There is a checkpatch warning

```
WARNING: Missing or malformed SPDX-License-Identifier tag in line 1
#25: FILE: arch/arm/boot/dts/xilinx/ni-bluefinlc.dts:1:
+/dts-v1/;
```

Something like [this](https://patchwork.ozlabs.org/project/linux-pci/patch/20190828135322.10370-1-kw@linux.com/) will fix?

chaitu236 (changes requested):

arch/arm/boot/dts/xilinx/ni-bluefinlc.dts
DeborahOoi96: thanks for the pointers!

DeborahOoi96:
@chaitu236 I added ply as import ply but for some reason it still complains no module named ply. I did rebase on the latest main, so I'm not sure why checkpatch.pl still fails. It's not even a patch to begin with, it's just a commit so I'm a little confused why run checkpatch. Any pointers will be appreciated :)

arch/arm/boot/dts/xilinx/ni-bluefinlc.dts
```diff
@@ -0,0 +1,138 @@
...
+	gpio_keys {
+		compatible = "gpio-keys";
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		/* Reset switch is on GPIO48 */
+		reset_sw@0 {
+			label = "reset_sw";
+			gpios = <&gpio0 48 1 /* GPIO_ACTIVE_LOW */>;
+			linux,code = <408>; /* KEY_RESTART */
```
bstreiff, line 40: if you #include <dt-bindings/gpio/gpio.h> and <dt-bindings/inxput/linux-event-codes.h> here, you should be able to use GPIO_ACTIVE_LOW in place of 1 and KEY_RESTART in place of 408.

(The old process that we had for building linux_kernel_bluefin didn't use the kernel's makefile process to build the device trees, it called `dtc` itself, and thus did not have access to those include files.)

(This can be filed as a cleanup work item to follow up on, since ni-bluefin.dts has the same problem. tbh most of this file could be refactored into a common dtsi for both ni-bluefin.dts and ni-bluefinlc.dts to include...)

scripts/spdxcheck.py
```diff
@@ -3,6 +3,7 @@
 # Copyright Thomas Gleixner <tglx@linutronix.de>
 
 from argparse import ArgumentParser
+import ply
```
bstreiff: NAK on this commit, we're already importing ply via `from ply`, I don't think this is correct.

I suspect the runner on which we're running checkpatch.pl is missing the `python3-ply` package; `pr-checks.yml` probably should make sure it's `apt install`ed as a build dependency. I don't see it among [the things that GitHub says are on the ubuntu-latest images](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md), so otherwise it'd only end up there if it were a dependency of something else in that image.

bstreiff (commented):
> It's not even a patch to begin with, it's just a commit so I'm a little confused why run checkpatch.

The commit is itself a patch. `git` is just a bunch of tools for e-mailing patches around. (Just with increasingly nebulous definitions of "e-mail", since git forges like github do a very good job of hiding that part.)

arch/arm/boot/dts/xilinx/ni-bluefinlc.dts
```diff
@@ -0,0 +1,138 @@
+// SPDX-License-Identifier: GPL-2.0
+/dts-v1/;
+ #include <dt-bindings/interrupt-controller/irq.h>
```
chaitu236: Remove space at beginning of line

DeborahOoi96: this space was introduced by checkpatch.pl --fix. should it not be there?

chaitu236: I don't see a space in other dts files in kernel. It might be a bug in `checkpatcch.pl --fix`.

scripts/spdxcheck.py
chaitu236: @DeborahOoi96 https://github.com/ni/linux/blob/nilrt/master/6.12/.github/workflows/pr-checks.yml#L23 - this is where we may want to install it.

DeborahOoi96: Thanks for the pointers, but now I'm getting more checkpatch.pl warnings that do not appear actionable.

MAINTAINERS warning: I don't think adding a DTS file should require updating MAINTAINERS? There is only one emerson entry there which is @chaitu236 's name. 

DT compatible warning ("ni,zynq"): The ni,* prefix is already defined in vendor-prefixes.yaml, so this looks like a false positive. 

At this point I’m not sure what changes are expected, and this is the main thing blocking progress on my side.
Could we unblock this or suggest a way to bypass/resolve these checks?

arch/arm/boot/dts/xilinx/ni-bluefinlc.dts
DeborahOoi96: I think I will create a separate user story / tech debt for this and resolve this comment together with other changes to ni-bluefin.dts.

DeborahOoi96: Tech debt created here:
https://dev.azure.com/ni/DevCentral/_workitems/edit/3824710/

scripts/spdxcheck.py
chaitu236: I think those warnings are ok to ignore.

chaitu236 (approved):

bstreiff (approved):

PR #276, opened 2026-06-03
dts: Add device tree for fielddaq devices

DeborahOoi96:
Device tree for FieldDAQ devices are for FD-11613, FD-11614 and FD-11637. There were other FieldDAQs that existed in bluefin/master/4.1 branch but those are discontinued now.
The changes here are to update the model name, NIDEVCODE, addition of ref100 and si5351c chip, and some LED changes. 
Otherwise, the rest of the device tree remains the same as Bluefin.

Testing:
When built together with these changes https://github.com/ni/meta-expio/pull/64, we were able to get ni-fielddaq.dtb generated.
image


chaitu236 (approved):

gratian:
Also cherry-picked in `nilrt/master/6.18` as commit https://github.com/ni/linux/commit/b575b5ca988279a7b771a37f527335c64930998a

PR #277, opened 2026-06-05
dts: Update devcodes for ni-fielddaq devices

DeborahOoi96:
These devices are not discontinued, we should add them back in.


gratian (approved):

gratian:
Also cherry-picked in `nilrt/master/6.18`.

PR #290, opened 2026-08-24
[ARM] Support FPGA, fixed IRQ mapping

chaitu236:
Support FPGA programming and enable fixed IRQ mapping.

WI: [AB#3990488](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3990488)

### Testing
- [x] Built BSI, safemode on ARM.
- [x] Boot tested on cRIO-9068, cRIO-9066.
- [x] FPGA can be programmed on cRIO-9068, cRIO-9066.
- [x] Verified network and serial console are accessible after FPGA is programmed on cRIO-9068.
- [x] Verified network is accessible after FPGA is programmed on cRIO-9066.
- [x] Verified fixed IRQ mappings work on cRIO-9068 by installing `ni-crio-driver-zynq-dkms`, running an IRQ test LV project and verifying irq count increases with `grep atomiczynq /proc/interrupts`.


chaitu236:
v3: Added `Assisted-by` tags. Retained original sign off tags from https://github.com/ni/linux/commit/dcf0b05acb5d06457730550c471aa643c9477029

drivers/irqchip/irq-gic.c
```diff
@@ -1160,6 +1160,60 @@ static const struct irq_domain_ops gic_irq_domain_hierarchy_ops = {
...
+	unsigned int irq;
+	int irq_base;
+
+	if (!IS_ENABLED(CONFIG_NI_ZYNQ_GIC_LEGACY_IRQDOMAIN) ||
+	    gic != &gic_data[0])
+		return irq_domain_create_linear(handle, gic_irqs,
+						&gic_irq_domain_hierarchy_ops,
+						gic);
+
+	irq_count = gic_irqs - NR_IRQS_LEGACY;
```
gratian, line 1188: Can this underflow if `gic_irqs < NR_IRQS_LEGACY`? Do we need a warn here just in case? i.e. `WARN_ON(gic_irqs <= NR_IRQS_LEGACY)`

drivers/irqchip/irq-gic.c
```diff
@@ -1160,6 +1160,60 @@ static const struct irq_domain_ops gic_irq_domain_hierarchy_ops = {
...
+	if (!domain) {
+		irq_free_descs(irq_base, irq_count);
+		return NULL;
+	}
+
+	for (irq = NR_IRQS_LEGACY; irq < gic_irqs; irq++) {
+		if (irq_find_mapping(domain, irq) != irq) {
+			pr_err("GIC: failed to establish fixed mapping for IRQ%u\n",
+			       irq);
+			return NULL;
```
gratian, line 1221: Do we need to deallocate anything here since it's an error case?

chaitu236:
v4: Fix checkpatch for 'Assisted-by' tags.

drivers/tty/serial/8250/8250_ni.c
```diff
@@ -77,8 +82,58 @@ struct ni16550_device_info {
...
+					unsigned int offset, u32 value)
+{
+}
+
+static int ni16550_fpga_notify(struct notifier_block *nb,
+			      unsigned long event, void *unused)
+{
+	struct ni16550_data *data = container_of(nb, struct ni16550_data,
+						 fpga_notifier);
+	struct uart_8250_port *uart = serial8250_get_port(data->line);
```
gratian, line 109: Copilot suggests a NULL check here `if (!uart) return NOTIFY_DONE` to avoid a possible race with `ni16550_remove()` which does remove the notifier before calling `serial8250_unregister_port()` but there is a small race window because notifier delivery is a global chain and the callback depends on data->line still pointing at a live port.

It's probably good cheap insurance to add the NULL check.

drivers/tty/serial/8250/8250_ni.c
```diff
@@ -77,8 +82,58 @@ struct ni16550_device_info {
...
+		}
+		break;
+	case FPGA_PERIPHERAL_UP:
+		if (data->fpga_suspended) {
+			serial8250_set_defaults(uart);
+			serial8250_resume_port(data->line);
+			data->fpga_suspended = false;
+		}
+		break;
+	case FPGA_PERIPHERAL_FAILED:
```
gratian, line 127: Do we need to handle anything for the failed case? Seems weird that we're not doing the suspend work. I guess the old commit didn't do much here either so maybe that's OK.

gratian (approved):
Looks OK overall, just a couple of nitpicks.

drivers/tty/serial/8250/8250_ni.c
chaitu236: It appears that `blocking_notifier_chain_unregister` in `ni16550_remove` waits until all callbacks are finished. And once notifier is unregistered, `ni16550_fpga_notify` should never be called so there should be no race.

drivers/tty/serial/8250/8250_ni.c
chaitu236: Opening fpgaperipheral causes FPGA_PERIPHERAL_DOWN which would have suspended the ports.

FPGA_PERIPHERAL_FAILED happens when closing fpgaperipheral by which point ports should have already been suspended. So I don't think we need to do anything here.

drivers/irqchip/irq-gic.c
chaitu236: Good point. Now returning NULL.

drivers/irqchip/irq-gic.c
chaitu236: Done

chaitu236:
v5: Address review feedback

PR #298, opened 2026-09-23
nati_zynq_defconfig: enable nftables backend

Vairamani-NI:
### Summary
Enable the nftables backend on the Zynq (cRIO ARM) defconfig so firewalld works on
NILRT ARM targets. `CONFIG_NF_TABLES` was unset, so firewalld cannot start on ARM
(`nft` reports "cache initialization failed: Invalid argument"; `nft add table inet`
returns "Operation not supported"). This enables nftables plus NAT/masquerade and
conntrack, matching what x86_64 already has.

- `CONFIG_NF_CONNTRACK`: `m` → `y`
- Add `NF_TABLES` (+ `inet`/`ipv4`/`ipv6`), `NF_NAT`/`NF_NAT_MASQUERADE`,
  `NF_REJECT_IPV4`/`IPV6`, `NFT_CT`/`NAT`/`MASQ`/`REJECT`/`REJECT_INET`/
  `COUNTER`/`LOG`/`LIMIT`/`FIB*`
- All built-in (`=y`) because the ARM NILRT image ships no netfilter kernel modules.

### Justification
[AB#3950888](https://dev.azure.com/ni/DevCentral/_workitems/edit/3950888)

### Testing
Validated on a cRIO-9068 (kernel 6.18) built with this config:
- `CONFIG_NF_TABLES=y`; `nft add table inet <t>` succeeds.
- firewalld 2.2.1 runs on the nftables backend: default-deny `public` zone with
  ssh/http/https + NI services, 24 NI service defs, and `firewall-cmd
  --add-masquerade` produces real nftables `nat_PREROUTING`/`nat_POSTROUTING`
  chains (inet-family NAT — the capability the 4.14 kernel lacked).

### Notes
- This defconfig is shared by the ARM runmode and safemode kernels, so it enables
  nftables for both.
- Validated on hardware (cRIO-9068, kernel 6.18):
  - Runmode: firewalld runs on the nftables backend — default-deny public zone,
    NI services, and NAT/masquerade all working. Also re-checked under a 256 MB
    RAM cap (mem=256M) with no change in behavior.
  - Safemode: ni-safemode-firewall enforces default-deny (ssh/http/https/nisvcloc
    open, other ports blocked).


Vairamani-NI:
@ni/rtos pls review this.

Vairamani-NI:
CI failed because the defconfig was hand-edited, not savedefconfig-generated. Regenerated it properly with make ARCH=arm savedefconfig and force-pushed; should pass now.

rajendra-desai-ni (commented):
You can remove `ARM` prefix from the commit message.
`nati_zynq_defconfig: enable nftables backend`

arch/arm/configs/nati_zynq_defconfig
```diff
@@ -71,7 +71,7 @@ CONFIG_INET6_ESP=m
 CONFIG_INET6_IPCOMP=m
 CONFIG_IPV6_MULTIPLE_TABLES=y
 CONFIG_NETFILTER=y
-CONFIG_NF_CONNTRACK=m
+CONFIG_NF_CONNTRACK=y
```
gratian: Does this have to be built-in? The commit message does not explain why.

For x64 we have it configured as a module: https://github.com/ni/linux/blob/4c7fe3ef67d6119d9fe1a8d56013b49e344d7cac/arch/x86/configs/nati_x86_64_defconfig#L120

Modules would be preferable unless there is a particular reason we need it built-in.

arch/arm/configs/nati_zynq_defconfig
```diff
@@ -85,6 +85,15 @@ CONFIG_NF_CONNTRACK_SANE=m
 CONFIG_NF_CONNTRACK_SIP=m
 CONFIG_NF_CONNTRACK_TFTP=m
 CONFIG_NF_CT_NETLINK=m
+CONFIG_NF_TABLES=y
```
gratian: Same as above. Check if things really need to be built-in.

gratian (commented):
Please check if the built-in (=y) options are necessary or if modules would work instead. User nati_x86_64_defconfig for reference.

Vairamani-NI:
> Please check if the built-in (=y) options are necessary or if modules would work instead. User nati_x86_64_defconfig for reference.

Switched everything to modules to match nati_x86_64_defconfig. NF_CONNTRACK stays =m, NF_TABLES and all NFT_* are now =m, with only NF_TABLES_INET=y (boolean sub-option). Verified the arm runmode rootfs and safemode initramfs already package the =m netfilter modules, so firewalld autoloads them the same way it does on x64. Regenerated with savedefconfig.

gratian (approved):

